PR #12, opened 2020-06-15
improve types

kotarella1110:


src/superstruct.ts
```diff
@@ -31,7 +20,7 @@ const parseErrorSchema = (
                       currentPath,
                       validateAllFieldCriteria,
                       previous,
-                      type,
+                      type!,
```
kotarella1110, line 23: @bluebill1049 `type` may be `undefined`. 
using `!` to ignore type errors, how should I fix this?

bluebill1049: maybe we should omit the type if it can be `undefined`? your thoughts?

src/superstruct.ts
```diff
@@ -41,7 +30,7 @@ const parseErrorSchema = (
                       type,
                       ...(validateAllFieldCriteria
                         ? {
-                            types: { [type]: message || true },
+                            types: { [type!]: message || true },
```
kotarella1110, line 33: Same as above

src/superstruct.ts
kotarella1110: @bluebill1049 If we omit type, I think you need to fix `FieldError` type as follows.

```diff
 export type FieldError = {
-  type: string;
+  type?: string;
   ref?: Ref;
   types?: MultipleFieldErrors;
   message?: Message;
   isManual?: boolean;
 }
```

I think it's better to use an empty string (`''`) to avoid this fix.

bluebill1049: oh that's right, let's leave as ''

kotarella1110: I will fix this!

kotarella1110:
Yes, I want to release a new patch!

bluebill1049:
make sure push the tag and leave a release note as well :) 👍

kotarella1110:
What about the version?
`1.0.0-beta.0` OK?

bluebill1049:
we can go v1 if you are confident.

kotarella1110:
I will release `0.0.4-beta.0` 👍

kotarella1110:
@bluebill1049 thanks!

Issue #7, opened 2020-05-27
Question: supporting json schema / ajv


joshyudeep:
First off, thanks for all the work y'all put into this. 

Are you guys planning to add a resolver for json schema / ajv? 
https://json-schema.org/specification.html
https://github.com/ajv-validator/ajv

bluebill1049:
That would be great, would you please to contribute a resolver for that? https://github.com/react-hook-form/react-hook-form-resolvers V6 will support different types of schema.

QuentinDejean:
Hey guys! Great to see a PR open for that one!

Can we hope to see it merged anytime soon?

TheDSCPL:
This would be awesome! I have a backend writter in Golang and all the JSON schemas are already defined. Being able to use the same JSON schema would be really good for making the code more concise and DRY. Also, the JSON schema is a [standard](http://json-schema.org/) (ajv is just an implementation for javascript) and it should have been one of the first validation methods.

bluebill1049:
anyone would like to take over that PR and address some of the comments?

TheDSCPL:
I'll read through the code and comments to see if I can help

TheDSCPL:
I've finished writing the code. It working with synchronous schemas and also hides the fact that it uses ajv behind the scenes. I called the new resolver `jsonSchema` because that's the actual model it's following. Ajv is an implementation detail.

I'll create a PR.

TheDSCPL:
I also added `@types/json-schema` as a dev dependency to add type safety to the schema input :grin:

qlaffont:
Any update on this ? I didn't see it anymore on ReadMe since rework :/

jorisre:
There is an opened PR https://github.com/react-hook-form/resolvers/pull/59 with some issues

bluebill1049:
Sorry, we are not going to support ajv for now. i know this will make some users angry, but the module ajv old and hard to integrate as far as I can tell. if you need to get it working, please build a custom resolver at your app level.

AntonOfTheWoods:
@bluebill1049 , do you have any further information on what you mean by "the module ajv old and hard to integrate"? ajv is a lot faster than most of the competitors, is still actively developed by lots of developers and has a lot of users, downloads and stars. I didn't look in detail at the code but if it is abstracted properly then I'm not sure I see why that would be a `wontfix`.

Or do you simply not want to support json schema at all?

bluebill1049:
[...]

PR #14, opened 2020-06-17
feat: Added initial implementation of ajvResolver

lburgess:
AJV's validation API is a little clunky with a few different ways to use it and no real normalisation (some times it throws an exception and other times it sets the errors on what should ideally be an immutable object).

Currently only supports $async schemas, but this is easy to append to any JSON schema so deemed a suitable trade off as an initial version and this supports all use cases, be that local, in-memory validation (sync or async) or asynchronous validation via a remote call.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@types/yup": "^0.29.3",
     "@typescript-eslint/eslint-plugin": "^3.2.0",
     "@typescript-eslint/parser": "^3.2.0",
+    "ajv": "^6.12.2",
     "eslint": "^7.2.0",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-prettier": "^3.1.4",
--- a/src/__snapshots__/ajv.test.ts.snap
+++ b/src/__snapshots__/ajv.test.ts.snap
@@ -1,0 +1,81 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ajvResolver using Ajv.compile should get errors for array fields 1`] = `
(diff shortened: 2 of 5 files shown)


bluebill1049:
awesome stuff! close https://github.com/react-hook-form/resolvers/issues/7

kotarella1110:
I'm busy, so I'll review this next week.
Please wait a little.

bluebill1049:
Agree, it's better to keep it consistent.

TheDSCPL:
I created a [PR](https://github.com/react-hook-form/resolvers/pull/50) for this. Fixed the errors and the need for async schemas.

Your PR handled the errors incorrectly because you didn't call `transformToNestObject` before returning the errors, you passed the `params` property from the ajv errors into the resolver's error params (which aren't the same) and you didn't set the `type` property of the resolver's error (in ajv errors it's the `keyword` property).
Also at least one of your tests (`using Ajv.compile should get values`) was wrong because the input data had the `age` property set to `'24'` (string) and the schema you defined was expecting an integer greater or equal to 0 in that property but the validation was still coming up without an error.

I forked your PR to reuse some of your code and give you credit for doing this first :smile: I had been meaning to contribute to the library with this implementation but I never got around to do it. I implemented a verification with ajv seperate from the `react-hook-form` which set the errors manually because the library didn't support ajv. This will come in super handy because I'll be able to use the same schema for validation in the backend and in the frontend without worrying about making any mistakes converting the schemas to another library's implementation.

jorisre:
Closing due to inactivity

PR #18, opened 2020-07-07
Feature/nope resolver

kysley:
Hey @bluebill1049, big fan ;) 

I made this to use on a new project of mine, and figured I should put up a pr.

Let me know what you guys think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,14 +75,14 @@ A simple and composable way to validate data in JavaScript (or TypeScript).
 [![npm](https://img.shields.io/bundlephobia/minzip/superstruct?style=for-the-badge)](https://bundlephobia.com/result?p=superstruct)
 
 ```typescript jsx
-import React from "react";
-import { useForm } from "react-hook-form";
-import { superstructResolver } from "@hookform/resolvers";
-import { struct } from "superstruct";
+import React from 'react';
+import { useForm } from 'react-hook-form';
+import { superstructResolver } from '@hookform/resolvers';
+import { struct } from 'superstruct';
 
 const schema = struct({
-  name: "string",
-  age: "number",
+  name: 'string',
+  age: 'number',
 });
 
 const App = () => {
@@ -108,10 +108,10 @@ The most powerful data validation library for JS.
 [![npm](https://img.shields.io/bundlephobia/minzip/@hapi/joi?style=for-the-badge)](https://bundlephobia.com/result?p=@hapi/joi)
 
 ```typescript jsx
-import React from "react";
-import { useForm } from "react-hook-form";
-import { joiResolver } from "@hookform/resolvers";
-import Joi from "@hapi/joi";
+import React from 'react';
+import { useForm } from 'react-hook-form';
+import { joiResolver } from '@hookform/resolvers';
+import Joi from '@hapi/joi';
 
 const schema = Joi.object({
   username: Joi.string().required(),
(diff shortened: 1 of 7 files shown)


bluebill1049:
oh wow! nice! thank you very much for this.

JeromeDeLeon:
This is looking great! However, I believe that this could also be put up on a website. The README is starting to be bloated with examples. I think it makes sense to include a list of examples and link them to it.

JeromeDeLeon:
Haven't seen any other library aside from the first three on the README. This kinda a great to experiment. Thank you. Really appreciated it.

src/nope.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  it('should return errors', async () => {
+    const data = {
+      id: '2',
+      title: 2,
+      tags: [2, 3],
+      author: {
+        id: 'test',
+      },
+    };
+    expect(await nopeResolver(Article)(data)).toMatchSnapshot();
```
keiya01, line 39: I think you should check error value. Because I think snapshot test should be used only when comparing value is difficult.

src/nope.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const parseNopeSchema = (error: any, validateAllFieldCriteria: boolean) => {
+  const errors = Object.keys(error);
+
+  return Array.isArray(errors)
+    ? errors.reduce((acc: Record<string, any>, fieldName) => {
+        return {
+          ...acc,
+          ...(fieldName
+            ? acc[fieldName] && validateAllFieldCriteria
```
keiya01, line 12: I think you should test here too. (the case `validateAllFieldCriteria` is false)

src/nope.ts
```diff
@@ -0,0 +1,61 @@
+import { appendErrors, transformToNestObject, Resolver } from 'react-hook-form';
+import NopeObject from 'nope-validator/lib/umd/NopeObject';
+
+const parseNopeSchema = (error: any, validateAllFieldCriteria: boolean) => {
+  const errors = Object.keys(error);
+
+  return Array.isArray(errors)
```
keiya01, line 7: I think `errors` is always `Array` type. Is it wrong?

keiya01:
@kysley 
Thank you very much for sending the PR🙏
I reviewed your code.

src/nope.test.ts
kysley: Nope! This is the signature:

```
export interface ShapeErrors {
    [key: string]: string;
}
```

src/nope.ts
kysley: Nope! This is the signature:

```ts
export interface ShapeErrors {
    [key: string]: string;
}
```

src/nope.ts
```diff
@@ -0,0 +1,61 @@
+import { appendErrors, transformToNestObject, Resolver } from 'react-hook-form';
+import NopeObject from 'nope-validator/lib/umd/NopeObject';
```
kotarella1110, line 2: Why import umd?

src/nope.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const parseNopeSchema = (error: any, validateAllFieldCriteria: boolean) => {
```
kotarella1110, line 4: Is this possible to avoid any?

```ts
import { ShapeErrors } from 'nope-validator/lib/umd/types';

const parseNopeSchema = (error: ShapeErrors, validateAllFieldCriteria: boolean) => {
```

kysley:
Hey guys, I'm new to using Typescript in more complex situations, and would love a little guidance here.

Any `NopePrimitive` can be validated, changing the return signature.
```js
Nope.string() //String, Number, Date, Boolean
  .test(a => (a === '42' ? undefined : 'Must be 42'))
  .validate('41'); 

// string => 'Must be 42'
```
If `NopeObject` has `shape({...})`, then `ShapeErrors` is returned.
```js
const schema = Nope.object().shape({
  name: Nope.string()
    .atMost(15)
    .required(),
  email: Nope.string()
    .email('Please provide a valid email')
    .required(),
});

const errors = schema.validate({
  name: 'Test',
  email: 'invalidemail',
});

// ShapeErrors => { email: 'Please provide a valid email', }
```
I think to keep consistent, we should require a `NopeObject` WITH the `shape` signature. But I am unsure how to write a type signature for this scenario.

src/nope.ts
kysley: It might be more complicated than casting it to `ShapeErrors`, please see my comment below!

bluebill1049:
shall we get this boy in and sort out the type afterword? ship it first :)

kysley:
> shall we get this boy in and sort out the type afterword? ship it first :)

There seems to be an issue with `nope`. It doesn't validate nested objects within a schema. I think we should wait? Since RHF users may have nested form fields.

bluebill1049:
Thanks @kysley for the update, let's wait for nope validator author to response on the above :)  🙏

README.md
```diff
@@ -133,6 +133,39 @@ const App = () => {
...
+
+A small, simple and fast JS validator. Like, wow thats fast. 🚀
+
+[![npm](https://img.shields.io/bundlephobia/minzip/nope-validator?style=for-the-badge)](https://bundlephobia.com/result?p=nope-validator)
+
+```typescript jsx
+import React from 'react';
+import { useForm } from 'react-hook-form';
+import { nopeResolver } from '@hookform/resolvers';
+import Nope from 'nope-resolver';
```
jgeschwendt, line 146: Should this import `nope-validator` instead?

firebluetom2:
I would love to see this library supported, would want to replace my yup with something smaller

bluebill1049:
> I would love to see this library supported, would want to replace my yup with something smaller

use Zod instead.

jorisre:
I close this one because the nope resolver is released

PR #31, opened 2020-07-18
add Lazy Type

Ryomasao:
Hi! I want to use Yup.lazy, but I get a type error.

https://codesandbox.io/s/lingering-frog-0bryw?file=/src/App.tsx
image

It worked fine when I changed the type to any, so I added the `Lazy` type to Schema.
Lazy seems to be the same as the Schema type.

```
// tslint:disable-next-line:no-empty-interface
export interface Lazy extends Schema<any> {}
```


bluebill1049:
Thanks @Ryomasao

PR #30, opened 2020-07-16
feat(context): inject context into yup context

Vandell63:
Hey there,
-> https://github.com/react-hook-form/resolvers/issues/29
Test + context injected
I prefer to don't dive test because i will make another mr for -> https://github.com/react-hook-form/resolvers/issues/26#issuecomment-659114319.
And then push test to 100%

Issue #26, opened 2020-07-13
Question: How to do multiple checks for password validation with Yup and RFH?


sngreg:
I am wondering if there is a way with yup and yupResolver with RHF to do something similar to this:

https://codesandbox.io/s/eager-bush-dfl5t

Basically using this code when a user starts typing into the password field you get separated booleans for each validation criteria which allows you to build an interactive component such as:

<img width="397" alt="Screen Shot 2020-07-13 at 7 40 26 PM" src="https://user-images.githubusercontent.com/6288003/87364121-c3b1a700-c540-11ea-9ee5-9cf8d86c89a4.png">

Code that I have been playing around with:

```
....
import { yupResolver } from '@hookform/resolvers';

const passwordSchema = yup.object().shape({
  password: yup
    .string()
    .required()
    .min(8)
    .matches(RegExp('(.*[a-z].*)'), 'Lowercase')
    .matches(RegExp('(.*[A-Z].*)'), 'Uppercase')
    .matches(RegExp('(.*\\d.*)'), 'Number')
    .matches(RegExp('[!@#$%^&*(),.?":{}|<>]'), 'Special'),
});

  const { register, handleSubmit, errors, formState, getValues } = useForm({
    resolver: yupResolver(
      passwordSchema
    ),
    criteriaMode: 'all',
    reValidateMode: 'onChange',
    mode: 'onChange',
  });
....
```

Which results in:

<img width="579" alt="Screen Shot 2020-07-13 at 7 47 39 PM" src="https://user-images.githubusercontent.com/6288003/87364478-c2cd4500-c541-11ea-94d7-836389a70ca8.png">

I was hoping to have types returned like the coding sample so I can flag the realtime critera.

Any help or things in yup would be super helpful.

Thanks!

bluebill1049:
we do actually support `validateAllFieldCriteria` https://github.com/react-hook-form/resolvers/blob/master/src/yup.ts otherwise it would be a bug if it's not working.

sngreg:
Thanks @bluebill1049  for the quick reply. Going to try that quickly now to see if that is what I am looking for. Will post here and close out if it works :)

sngreg:
I am pretty new to RFH and Yup looking at the types, I am not seeing `validateAllFieldCriteria`

```
// Not Correct
 resolver: yupResolver(
      passwordSchema,
      { validateAllFieldCriteria: true }
    ),```

This is what it is expecting:

[...]

PR #32, opened 2020-07-19
Support multiple errors messages

Vandell63:
- [x] 100% test coverage for the yup resolver
- [x] Now errors have a types props with a Record key type and for values an array of error messages, if `criteriaMode: "all"` 

**At the same time:** 
- [x] Fix the abort early issue. It does not work because it let a yup resolver with an `inner error` of an empty array. So it always return any errors. Now we check if inner errors is an empty array or not. If no please return only one error at a time. Test has been done for this. it could help resolve issue from -> https://github.com/react-hook-form/resolvers/pull/30#issuecomment-660380652 where yup validation return anything.
- [x] Optimise the validate all field criteria props in the yup resolver script. The yup error path is always defined so i remove some useless code logic ( if i don't make a mistake )
- [x] The new props `types` is always defined it can be an empty array too.
- [x] Still compatible with the ErrorMessage component. 
- [x] Add more types to the resolver script -> to be check im not an expert. I used types from rhf dist i think maybe it is not the right way :/

**Improvements which can be done into**

1.  For an array, If we remove the `transformToNestObject` we will have at the same time array errors and all field errors inside this array.
Cons: errors duplication : 
```
foo[0].test: {
      message: ...,
      type: ...,
      types: [ ] | {}
},
foo: [
  { 
    test: {
      message: ...,
      type: ...,
      types: [ ] | {}
    }
  },
  ...
]
```
Pros: it can be interesting to have it to get at the same time array errors and fields errors in.
We could maybe do : `validateAllFieldCriteria ? call transformToNestObject `
You can have example of use case here -> https://github.com/react-hook-form/resolvers/issues/26#issuecomment-660498288
From issue -> https://github.com/react-hook-form/resolvers/issues/26

2. On first try to rhf we don't know where to place the context and where will be the changes !? We should remove the context option from the yup options by his types. It come in trouble when you expect your form validation to be used from cached context or from yup context options. From my issue -> https://github.com/react-hook-form/react-hook-form/issues/2222. We could do this to force user to always use the context manage by rhf:
```
export const yupResolver = <TFieldValues extends Record<string, any>>(
  schema: Yup.ObjectSchema | Yup.Lazy,
  options: Omit<Yup.ValidateOptions, 'context'> = { // OMIT THE CONTEXT
    abortEarly: false,
  },
```

it could help issue like -> https://github.com/react-hook-form/resolvers/pull/30#issuecomment-660380652

### DIFF
--- a/src/__snapshots__/yup.test.ts.snap
+++ b/src/__snapshots__/yup.test.ts.snap
@@ -6,26 +6,42 @@ Object {
     "age": Object {
       "message": "age is a required field",
       "type": "required",
+      "types": Object {
+        "required": "age is a required field",
+      },
     },
     "name": Object {
       "message": "name is a required field",
       "type": "required",
+      "types": Object {
+        "min": "name is a min field",
+      },
     },
   },
   "values": Object {},
 }
 `;
 
-exports[`yupResolver should get errors 1`] = `
+exports[`yupResolver errors should get errors with validate all criteria fields 1`] = `
 Object {
   "errors": Object {
     "age": Object {
       "message": "age must be a \`number\` type, but the final value was: \`NaN\` (cast from the value \`\\"test\\"\`).",
       "type": "typeError",
+      "types": Object {
+        "typeError": Array [
+          "age must be a \`number\` type, but the final value was: \`NaN\` (cast from the value \`\\"test\\"\`).",
+        ],
+      },
     },
     "createdOn": Object {
       "message": "createdOn must be a \`date\` type, but the final value was: \`Invalid Date\`.",
       "type": "typeError",
+      "types": Object {
+        "typeError": Array [
+          "createdOn must be a \`date\` type, but the final value was: \`Invalid Date\`.",
+        ],
+      },
     },
     "foo": Array [
       Object {
@@ -33,9 +49,75 @@ Object {
           "message": "foo[0].loose must be a \`boolean\` type, but the final value was: \`null\`.
  If \\"null\\" is intended as an empty value be sure to mark the schema as \`.nullable()\`",
           "type": "typeError",
+          "types": Object {
+            "typeError": Array [
+              "foo[0].loose must be a \`boolean\` type, but the final value was: \`null\`.
+ If \\"null\\" is intended as an empty value be sure to mark the schema as \`.nullable()\`",
+            ],
+          },
         },
       },
     ],
+    "password": Object {
+      "message": "password is a required field",
+      "type": "required",
+      "types": Object {
+        "matches": Array [
+          "Special",
+          "Number",
+          "Uppercase",
+          "Lowercase",
+        ],
+        "min": Array [
+          "password must be at least 8 characters",
+        ],
+        "required": Array [
+          "password is a required field",
+        ],
+      },
+    },
+  },
+  "values": Object {},
+}
+`;
+
+exports[`yupResolver errors should get errors without validate all criteria fields 1`] = `
+Object {
+  "errors": Object {
+    "age": Object {
+      "message": "age must be a \`number\` type, but the final value was: \`NaN\` (cast from the value \`\\"test\\"\`).",
+      "type": "typeError",
+      "types": Object {
+        "typeError": "age must be a \`number\` type, but the final value was: \`NaN\` (cast from the value \`\\"test\\"\`).",
+      },
+    },
+    "createdOn": Object {
+      "message": "createdOn must be a \`date\` type, but the final value was: \`Invalid Date\`.",
+      "type": "typeError",
+      "types": Object {
+        "typeError": "createdOn must be a \`date\` type, but the final value was: \`Invalid Date\`.",
+      },
+    },
+    "foo": Array [
+      Object {
+        "loose": Object {
+          "message": "foo[0].loose must be a \`boolean\` type, but the final value was: \`null\`.
+ If \\"null\\" is intended as an empty value be sure to mark the schema as \`.nullable()\`",
+          "type": "typeError",
+          "types": Object {
+            "typeError": "foo[0].loose must be a \`boolean\` type, but the final value was: \`null\`.
+ If \\"null\\" is intended as an empty value be sure to mark the schema as \`.nullable()\`",
+          },
+        },
+      },
+    ],
+    "password": Object {
+      "message": "password is a required field",
+      "type": "required",
+      "types": Object {
+        "required": "password is a required field",
+      },
+    },
   },
   "values": Object {},
 }
--- a/src/__snapshots__/yup.test.ts.snap
+++ b/src/__snapshots__/yup.test.ts.snap
@@ -33,19 +33,15 @@ Object {
         "typeError": "createdOn must be a \`date\` type, but the final value was: \`Invalid Date\`.",
       },
     },
-    "foo": Array [
-      Object {
-        "loose": Object {
-          "message": "foo[0].loose must be a \`boolean\` type, but the final value was: \`null\`.
+    "foo[0].loose": Object {
+      "message": "foo[0].loose must be a \`boolean\` type, but the final value was: \`null\`.
  If \\"null\\" is intended as an empty value be sure to mark the schema as \`.nullable()\`",
-          "type": "typeError",
-          "types": Object {
-            "typeError": "foo[0].loose must be a \`boolean\` type, but the final value was: \`null\`.
+      "type": "typeError",
+      "types": Object {
+        "typeError": "foo[0].loose must be a \`boolean\` type, but the final value was: \`null\`.
  If \\"null\\" is intended as an empty value be sure to mark the schema as \`.nullable()\`",
-          },
-        },
       },
-    ],
+    },
     "password": Object {
       "message": "password is a required field",
       "type": "required",
--- a/src/yup.test.ts
+++ b/src/yup.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import * as yup from 'yup';
 import { yupResolver } from './yup';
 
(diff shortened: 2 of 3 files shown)


Vandell63:
@bluebill1049  Test are failing because i am starting from the master and nope from my previous pull request. Should you want me to rebase from my previous PR to make test pass ? Or should i wait your new tag release ?

bluebill1049:
oh man! thanks very much for your amazing contribution, I will take a look at the PR. 🚀

bluebill1049:
Thanks @pmaier983 I will take a look at the PR tonight as well. solid effort @Vandell63

src/yup.test.ts
```diff
@@ -72,20 +81,101 @@ describe('yupResolver', () => {
...
+            ],
+            "min": Array [
+              "password must be at least 8 characters",
+            ],
+            "required": Array [
+              "password is a required field",
+            ],
+          }
+        `);
+    });
```
kotarella1110, line 132: The structure I had in mind was:

```ts
expect(resolve.errors['password'].types).toMatchInlineSnapshot(`
  Object {
    "matches": Array [
      "Special",
      "Number",
      "Uppercase",
      "Lowercase",
    ],
    "min": "password must be at least 8 characters",
    "required": "password is a required field",
  }
`);
```

We should discuss whether all should be array message when `validateAllFieldCriteria` is true.

bluebill1049: Thanks, @kotarella1110. Above is a breaking change, we should stick with 

```
expect(resolve.errors['password'].types).toMatchInlineSnapshot(`
  Object {
    "matches": Array [
      "Special",
      "Number",
      "Uppercase",
      "Lowercase",
    ],
    "min": "password must be at least 8 characters",
    "required": "password is a required field",
  }
`);
```

kotarella1110: @bluebill1049 I agree with you 👍

Vandell63:
> Thanks @pmaier983 I will take a look at the PR tonight as well. solid effort @Vandell63

As much as the time you gave me when i was in trouble

src/yup.test.ts
Vandell63: Yes, i was thinking if we keep this mix structure, i'm afraid that we have to handle the case where `messages` can be an array or a string in the code for every property in the schema. This is why i keep every messages in array format when `validateAllCriteria` is set to true. But you are in better place to take the right decision here :).

bluebill1049: Thanks very much for understanding and putting this PR together!

src/__snapshots__/yup.test.ts.snap
```diff
@@ -6,36 +6,118 @@ Object {
...
+      "types": Object {
+        "required": "age is a required field",
+      },
     },
     "name": Object {
       "message": "name is a required field",
       "type": "required",
+      "types": Object {
+        "min": "name is a min field",
+      },
```
kotarella1110, line 18: Same as above

kotarella1110:
@Vandell63 I have created a PR (https://github.com/Vandell63/resolvers/pull/1) to take my feedback comment above.
Please merge if no problem :pray:
After merging this PR, you are free to make changes.

Vandell63:
@kotarella1110  sry i was thinking that @bluebill1049  will make some change before. So if i understood, you want the props `types` to be undefined if `validateAllFieldCriteria` is set to false. To resume :
1. Can i keep the yup resolver script that i made and then change it to match the expected test result ? Or should i remove all my previous yup resolver change ?
2. Should i keep the mixed `array` and/or `string` types if `validateAllFieldCriteria` is set to true from your review @kotarella1110 ?
3. [Improvement] Should i remove the script `transformToNestObject` if `validateAllFieldCriteria` is set to false ?
4. [Improvement] Should i force the user to inject his context in the rhf options instead of yup options with the Omit type ?

kotarella1110:
> 1. Should i keep the yup resolver script that i made and then change it to match Your expected test result ? Or should i remove all my previous yup resolver changes and keep yours instead  ?

I will commit the No.5 fix to the PR I created.
`Array.isArray(error.inner) && error.inner.length` is `true`
You can then merge.

> 2. Should i keep the mixed `array` and/or `string` types if `validateAllFieldCriteria` is set to true from your review @kotarella1110 ?

yes. I think we should keep mixed types.

> 3. [Improvement] Should i remove the script `transformToNestObject` if `validateAllFieldCriteria` is set to `true` to get all fields error even if array have some ?

What do @bluebill1049 think about this ?

> 4. [Improvement] Should i force the user to inject his context in the rhf options instead of yup options with the Omit type ?

This is under consideration. Please give me some time.

> 5. Should i re-add the script code to
>    `Fix the abort early issue. It does not work because it let a yup resolver with an inner error of an empty array. So it always return any errors. Now we check if inner errors is an empty array or not`
>    By checking if
>    `Array.isArray(error.inner) && error.inner.length` is `true`

we should fix this 👍

bluebill1049:
> 3\. [Improvement] Should i remove the script `transformToNestObject` if `validateAllFieldCriteria` is set to `true` to get all fields error even if array have some ?

Yes totally 👍

bluebill1049:
Thanks @kotarella1110 for the review, sorry i was busy and haven't look close enough, i will make sure to do that tonight.

src/yup.ts
```diff
@@ -1,43 +1,44 @@
...
-                      : {}),
-                  },
-                }
-            : {}),
-        }),
+): DeepMap<TFieldValues, FieldError> =>
+  Array.isArray(error.inner) && error.inner.length
+    ? error.inner.reduce<FieldErrors<TFieldValues>>(
+        (previous, { path, message, type }) => {
+          const previousPath = previous[path] as FieldErrors | undefined;
```
bluebill1049, line 13: I think you can use this variable inline (save few bytes)

src/yup.ts
```diff
@@ -1,43 +1,44 @@
...
+            : {};
+          return {
+            ...previous,
+            [path]: {
+              ...(previous[path] || {
+                message,
+                type,
+                types: previousTypes,
+              }),
+              types: {
```
bluebill1049, line 25: `types` should be optional.

kotarella1110:
@Vandell63 @bluebill1049 please review this PR (https://github.com/Vandell63/resolvers/pull/1).

Vandell63:
> @Vandell63 @bluebill1049 please review this PR ([Vandell63#1](https://github.com/Vandell63/resolvers/pull/1)).

Done

Vandell63:
Thank @bluebill1049  @kotarella1110   merge done.
I will work on https://github.com/react-hook-form/resolvers/pull/32#issuecomment-661610874 and make the test for.
And I wait for the point `4` with @kotarella1110  message : `This is under consideration. Please give me some time.`

Vandell63:
@bluebill1049  @kotarella1110  Oh i maybe forget this. In yup errors the path seems to be never `undefined` so should i remove path logic from the code. Or should i keep it and make a test on it ?

Vandell63:
I finally added test and https://github.com/react-hook-form/resolvers/pull/32#issuecomment-661610874 logic.
Pending for my previous messages.

bluebill1049:
> 1. [Improvement] Should i force the user to inject his context in the rhf options instead of yup options with the Omit type ?

I think it would be better to centralize the `context` at useForm, that way our API is consistent through other resolvers as well. what do you think @kotarella1110?

bluebill1049:
> @bluebill1049 @kotarella1110 Oh i maybe forget this. In yup errors the path seems to be never `undefined` so should i remove path logic from the code. Or should i keep it and make a test on it ?

actually, from memory they do get `undefined` in certain circumstance, I can't remember when :(

kotarella1110:
> > 1. [Improvement] Should i force the user to inject his context in the rhf options instead of yup options with the Omit type ?
> 
> I think it would be better to centralize the `context` at useForm, that way our API is consistent through other resolvers as well. what do you think @kotarella1110?

@bluebill1049 Totally agree 👍 
@Vandell63 Let's remove the context option from yupResolver option.

Vandell63:
So @bluebill1049  @kotarella1110  should you prefer:
1. add the Omit typescript 
2. `delete options.context`
3. Add a console warn to the user if context is passed ( for non typescript dev ) to yup options on non-production build ?

It can be all of them. 
I let you me know what you prefer and then : 
- I add the missing test statement if yup path is undefined.
- Refactor context to be only the rhf context based on the previous selected options.
- Test the logic based on the previous selected options.

kotarella1110:
> https://github.com/react-hook-form/resolvers/pull/32#issuecomment-661855617

Does option 2 mean this?

```diff
 values:  await schema.validate(values, {
-   context,
    ...options,
+   context,
  }) as any,
```

I prefer all of them.
Let's wait for @bluebill1049 's opinion.

Vandell63:
Oh sorry @kotarella1110 , no in the code it could be 

```
try {
  if(options.context){
    console.warn("please use rhf context option instead")
    delete options.context
  }
  process()
} ...
```
Remove context property from yup option if user is not in typescript it could still add context props into. Just to be sure.

bluebill1049:
shall we do 1 and 3?

for TS user just a type
for JS user just a console.warn for dev only?

```
if (process.env.NODE_ENV ==='development') {
 console.warn('xxx');
}
```

kotarella1110:
@bluebill1049 I agree 👍 
@Vandell63 Also, we prefer the `context` passed from useForm over the `options.context` of yupResolver, so I think it should be fixed as I described in this https://github.com/react-hook-form/resolvers/pull/32#issuecomment-661887064.

Vandell63:
@bluebill1049  @kotarella1110  ok for me i will propose you a final commit with test as soon as i can .

bluebill1049:
> @bluebill1049 @kotarella1110 ok for me i will propose you a final commit with test as soon as i can .

Thank you very much. no rush whenever you are free.

Vandell63:
@bluebill1049  a pleasure for me no worry

Vandell63:
@bluebill1049  @kotarella1110  there we go.
- 100% test coverage
- Remove context type from the yup options
- Add warn log if yup context is defined
Maybe you will have to change the message, im bit crappy in english :/ .

bluebill1049:
Let's go! thanks @kotarella1110 's review and @Vandell63's hard work! do you have a twitter account? let's tweet this new feature.

Vandell63:
Yess it was so cool thx . let's go this new fresh twitter -> https://twitter.com/Paul_Vandell.
Thank you again for your high availability @bluebill1049  @kotarella1110

sngreg:
@Vandell63 @bluebill1049 @kotarella1110  - THANK you so much this code looks great. Sorry I went MIA for a couple of days, I was out of the office on vacation. 

This feature is a GREAT add to this repo!

PR #40, opened 2020-08-10
change target ES2017 to es5

### LINKED ISSUE #35: IE11 crashing with v0.1.0 and react-hook-form v6.08
It reports syntax error on this line:

`const parseErrorSchema = (error, validateAllFieldCriteria) => ...`

which is the code of the resolvers. I assume this needs some transpiling since IE doesn't know about const keyword?

kotarella1110:
close #35

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "sourceMap": true,
-    "module": "es2015",
-    "target": "es2017",
+    "module": "esnext",
+    "target": "es5",
     "moduleResolution": "node",
     "outDir": "./dist",
     "jsx": "react",


kotarella1110:
@bluebill1049 Since the target is changed to es5, the size after build will increase, but I think it can be used with IE11.

bluebill1049:
Thanks very much @kotarella1110 🙏 bloody IE...

kotarella1110:
@bluebill1049 I think this fix should be applied to @hookform packages as well. What do you think?

bluebill1049:
I agree @kotarella1110 🙏

bluebill1049:
until we stop supporting ie11.

kotarella1110:
@bluebill1049 yea. I will fix 👍

dylanhotfire:
Are you able to publish a release for this? It looks like index.cjs.developmenet.js has a compiled version of parseErrorSchema for the yup resolver, with a legitimate syntax error in IE and I'm wondering if this will fix it.

bluebill1049:
@dylanhotfire will release a beta for you to test.

bluebill1049:
`0.0.7-beta.1`

dylanhotfire:
Works like a charm. Thanks!

bluebill1049:
hey @kotarella1110 we may want to pause rolling this out. it does increase the package size quite a lot 👍 (probably except devtool package)

https://bundlephobia.com/result?p=@hookform/resolvers@0.0.7-beta.1

bluebill1049:
do we have the option to build an IE 11 build as well?

kotarella1110:
@bluebill1049 I think we need to build for IE11 like the core lib.

PR #45, opened 2020-08-19
fix #43 with yup.test

### LINKED ISSUE #43: Yup resolver problem with test function
Hello, 

It seems there is a problem with the `yupResolver`, when there is a schema with [`test` function](https://github.com/jquense/yup#mixedtestname-string-message-string--function-test-function-schema).

This a CodeSandbox to reproduce the problem: https://codesandbox.io/s/yup-resolver-problem-with-yup-test-function-16r4d?file=/src/App.js

The form example has 3 inputs. The age is required, only here to illustrate that individual field can throw, as expected, an error.
But the `test` function, which checks if email or name are present, does not throw a ValidationError.

The `onSubmit` method shows that there is no error and that the data is an empty object. 😢

bluebill1049:


bluebill1049:
no worries at all :) thanks for reviewing.

PR #48, opened 2020-08-26
improve rollup config to support IE11

### LINKED ISSUE #42: IE11 Support
This package does not work with IE11. The react-hook-form [latest release](https://github.com/react-hook-form/react-hook-form/pull/2542) looks like they're making efforts to support IE11 until it's unsupported, are we able to release a transpiled version of this package as well?

kotarella1110:
Usage:

```ts
import { yupResolver } from '@hookform/resolvers/dist/index.ie11.js';
```

if using TS:

create declaration file (`.d.ts`).

```ts
// global.d.ts
declare module '@hookform/resolvers/dist/index.ie11.js' {
    export * from '@hookform/resolvers';
}
```

if using webpack:

Note: not neeed to create declaration file if using TS.

```ts
import { yupResolver } from '@hookform/resolvers';
```

```js
// webpack.config.js
resolve: {
  alias: {
    "@hookform/resolvers": "@hookform/resolvers/dist/index.ie11.js"
  }
}
```

close #42
related https://github.com/react-hook-form/resolvers/pull/40

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -38,6 +38,14 @@ const options = [
     env: 'production',
     input: pkg.source,
   },
+  {
+    name,
+    umdName,
+    format: 'esm',
+    formatName: 'ie11',
+    input: pkg.source,
+    tsconfig: './tsconfig.ie11.json',
+  },
 ];
 
 export default options.map((option) => createRollupConfig(option));
--- a/rollup/createRollupConfig.js
+++ b/rollup/createRollupConfig.js
@@ -12,7 +12,7 @@ import { safePackageName } from './safePackageName';
 import { pascalcase } from './pascalcase';
 import pkg from '../package.json';
 
-export function createRollupConfig(options) {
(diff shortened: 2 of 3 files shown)


rollup.config.js
```diff
@@ -38,6 +38,14 @@ const options = [
     env: 'production',
     input: pkg.source,
   },
+  {
+    name,
+    umdName,
+    format: 'esm',
```
kotarella1110, line 44: Perhaps most users use webpack to bundle their app.
I think format may be ES module instead of CommonJS from the viewpoint of Tree Sharking.

IE11 support has been started in core lib with the following PR:

https://github.com/react-hook-form/react-hook-form/pull/235/files#diff-4dc771ce10b5894191688acc22e256b3

The core lib format is CommonJS, but I think this should also be fixed.

bluebill1049 (approved):
awesome! thanks, @kotarella1110 will build a beta version with this for testing as well.

PR #49, opened 2020-08-30
import types from core lib

kotarella1110:
related https://github.com/react-hook-form/react-hook-form/pull/2733


bluebill1049:
Thank you @kotarella1110 👍

Issue #20, opened 2020-07-08
Using a resolver for one of the validation libraries requires installation of other validation libraries.


avin-kavish:
I am trying to use yup resolver and it is telling me that package superstruct is missing.

bluebill1049:
which version? and screenshot?

avin-kavish:
`"@hookform/resolvers": "0.0.5"`

I just copied the yup resolver code and bypassed it for now, will post a screenshot when I can. It's happening with CRA + Typescript.

The error is basically the types for `superstruct` cannot be found. Once I installed 'superstruct', there was another error about generic parameters not being passed to superstruct type.

Basically, this type requires at least one generic parameter,

https://github.com/ianstormtaylor/superstruct/blob/01bdd37aef3357ff6296bf1827737ee11719e4b5/src/struct.ts#L9

But here there is none,

https://github.com/react-hook-form/resolvers/blob/ff82c5f442ee0b3a80da989ca36bf984b9d10404/src/superstruct.ts#L46

Will get back to you with screenshots.

bluebill1049:
Do you have any solution in mind?

qventura:
Why not having one package per resolver? 

As far as I can tell, this (https://github.com/react-hook-form/resolvers/blob/master/src/yup.ts) has no dependency to other module form the package.

This is actually how I fixed it in our project, just by extracting the content of `yup.ts` in one of our module.

bluebill1049:
I am seeking some evidence for this issue first before making any adjustments or changes.

avin-kavish:
This is a minimal reproduction: https://github.com/avin-kavish/cra-ts-hook-form-resolver-issue

![image](https://user-images.githubusercontent.com/48435155/86987245-34f50100-c1b3-11ea-995b-54006726b582.png)

bluebill1049:
Thanks @avin-kavish we can definitely try that.

bluebill1049:
I just tried your repo, if you have installed the dependency it's working as expected. 

<img width="549" alt="Screen Shot 2020-07-09 at 5 14 31 pm" src="https://user-images.githubusercontent.com/10513364/87008824-a8286280-c207-11ea-886d-4094363abd20.png">

bluebill1049:
Is this still an issue?

dev-araisan:
I have same issue in yupResolver (v0.0.6).
And I do not want to include other type definition packages.

[...]

PR #50, opened 2020-09-04
Feature/ajv resolver

TheDSCPL:
PR created due to [this](https://github.com/react-hook-form/resolvers/issues/7#issuecomment-687142903) by forking [this PR](https://github.com/react-hook-form/resolvers/pull/14).


TheDSCPL:
PR created due to [this](https://github.com/react-hook-form/resolvers/issues/7#issuecomment-687142903) by forking [this PR](https://github.com/react-hook-form/resolvers/pull/14).

TheDSCPL:
Made some changes:

- added tests for async schemas
- added tests to get more coverage
- removed some dead code
- added the ajv options to the resolver's options for more advanced use cases
- fixed a typo in the yup's section on the README.md file
- added alts to README.md's images

TheDSCPL:
Changed the TS config so now this PR passes the CI build

TheDSCPL:
Changed it back to the way I think it should be. I don't get why the CI build fails. It's the same configuration as in my computer and it works fine here. It's like the change to `tsconfig.json` (esModuleInterop) that I committed doesn't affect the CI build. Importing ajv needs that TS compiler option.

bluebill1049:
Thanks very much @TheDSCPL for the PR, I will take a look at it.

README.md
```diff
@@ -133,26 +133,73 @@ const App = () => {
...
+### [Json Schema](http://json-schema.org/)
+
+The most standard way to validate JSON (implemented by [ajv](https://github.com/ajv-validator/ajv))
+
+[![npm](https://img.shields.io/bundlephobia/minzip/ajv@6.12.4?style=for-the-badge)](https://bundlephobia.com/result?p=ajv@6.12.4)
+
+```typescript jsx
+import React from 'react';
+import { useForm } from 'react-hook-form';
+import { JSONSchema, jsonSchemaResolver } from '@hookform/resolvers';
```
bluebill1049, line 145: User should install 
`import Ajv, { ErrorObject, Options } from 'ajv';` at their app level, not within `resolves`.

src/jsonSchema.ts
```diff
@@ -0,0 +1,112 @@
...
+  options: JsonSchemaOptions = {},
+): Resolver<TFieldValues> => {
+  if (!validationSchema || typeof validationSchema !== 'object') {
+    throw new Error('Invalid AJV schema or validation function');
+  }
+
+  const async =
+    options.ajvOptions?.async === true ||
+    [true, 'true'].includes((validationSchema as any).$async);
+  const ajv = new Ajv({
```
bluebill1049, line 75: This should leave at the user's app-level, not within this lib, this lib only invoke the API and return the values/errors.

README.md
TheDSCPL: Ok! Of course.

src/jsonSchema.ts
TheDSCPL: Perfect. Then it's simpler and less prone to errors. I'll change these things. Btw, do you have any idea as to why the build fails in the CI?

README.md
bluebill1049: Thanks 👍

src/jsonSchema.ts
bluebill1049: no idea just yet :(

bluebill1049: I will take a closer look after :)

README.md
TheDSCPL: Wait! `JSONSchema` is not from `ajv`.It's a type that I created. It need to be imported from the resolvers

TheDSCPL: If the user want, they can also just pass the schema object. As long as it's correct, it'll pass the type check. This type is for users who want to statically type the schema object on declaration (and get auto completions from the IDE).

src/jsonSchema.ts
TheDSCPL: Made a new commit to remove the async logic

README.md
TheDSCPL: As I've said somewhere in these conversations, the fact that ajv is used is just an implementation detail. The user will just pass the schema (hence the schema type (which is not defined in ajv)) and doesn't need to import ajv.

TheDSCPL:
@bluebill1049 ?

bluebill1049:
I will take a look this weekend. 👍

README.md
bluebill1049: we already having an issue with types in the lib, can we please above export external type from this lib? https://github.com/react-hook-form/resolvers/issues/20

src/jsonSchema.ts
```diff
@@ -0,0 +1,98 @@
+import { appendErrors, Resolver, transformToNestObject } from 'react-hook-form';
+//import * as Ajv from 'ajv';
```
bluebill1049, line 2:
```suggestion
```

src/jsonSchema.ts
```diff
@@ -0,0 +1,98 @@
...
+import Ajv, { ErrorObject, Options } from 'ajv';
+import type { JSONSchema4, JSONSchema6, JSONSchema7 } from 'json-schema';
+import {
+  FieldError,
+  FieldErrors,
+  ResolverError,
+  ResolverSuccess,
+} from 'react-hook-form/dist/types/form';
+
+export type JSONSchema = JSONSchema4 | JSONSchema6 | JSONSchema7;
```
bluebill1049, line 12: I am not family with json-schema, what's 4 6 7?

src/jsonSchema.ts
```diff
@@ -0,0 +1,98 @@
...
+    Options,
+    'allErrors' | 'async' | 'validateSchema' | 'transpile'
+  >;
+}
+
+export const jsonSchemaResolver = <TFieldValues extends Record<string, any>>(
+  validationSchema: JSONSchema,
+  options: JsonSchemaOptions = {},
+): Resolver<TFieldValues> => {
+  if (!validationSchema || typeof validationSchema !== 'object') {
```
bluebill1049, line 69:
```suggestion
  if (typeof validationSchema !== 'object') {
```

don't think we need `!validationSchema` check here.

src/jsonSchema.ts
```diff
@@ -0,0 +1,98 @@
...
+  validationError: Array<ErrorObject> | null | undefined,
+  validateAllFieldCriteria: boolean,
+): FieldErrors<TFieldValues> =>
+  Array.isArray(validationError)
+    ? validationError.reduce(
+        (
+          previous: FieldErrors<TFieldValues>,
+          { dataPath, keyword, message },
+        ) => {
+          const path = dataPath.replace(/\//g, '.').replace(/^\./, '');
```
bluebill1049, line 24: I probably need to run it to understand better, so their path shape is different?

if so perhaps it's better to move this function outside the scope of the `reduce`.

src/jsonSchema.ts
```diff
@@ -0,0 +1,98 @@
...
+          };
+        },
+        {},
+      )
+    : {};
+
+export interface JsonSchemaOptions {
+  ajvOptions?: Omit<
+    Options,
+    'allErrors' | 'async' | 'validateSchema' | 'transpile'
```
bluebill1049, line 61: is `allErrors` means to return all the errors? that probably suitable for criticiaMode: 'all' at useForm.

bluebill1049:
Screen Shot 2020-09-12 at 8 31 16 am

looks like the build failed above.

bluebill1049:
if you need this urgently, i would recommend to publish a version of your own while we going through this, much appreciated your contribution.

src/jsonSchema.ts
TheDSCPL: Different standard versions. All of them are valid.

QuentinDejean:
Thanks a lot @TheDSCPL to get this PR going and @bluebill1049 to review it in such a timely manner, it's awesome to see progress on that space! ❤️

src/jsonSchema.ts
TheDSCPL: I believe you mean `criteriaMode`.

As per [AJV's documentation](https://github.com/ajv-validator/ajv#validation-and-reporting-options):

> - _allErrors_: check all rules collecting all errors. Default is to return after the first error.

You're right. That's exactly it. I didn't remember that option on `useForm`. Anyway, how do I even get that useForm option from the resolver? I thought about the `context` (second argument of the resolver) but it's undefined.

README.md
TheDSCPL: Ok, I read #20. I actually had already thought about that. I think this needs a restructuring.
I'd suggest having `@hookform/resolvers` defining the generic structure for validation resolvers, `@hookform/yupResolver` for the yup implementation, `@hookform/jsonSchemaResolver` for the JSON schema implementation, ... If all the implementations are under the same package then you'd have to install all the validation libraries as dependencies of this dependency just to use one of the implementations.

TheDSCPL: So for now all the dependencies of each resolver implementation needs to be in the `package.json` so this is correct. Nevertheless it should be changed to different packages. One for the generic resolver logic (for custom resolvers, for example) and one for each other implementation in order to isolate the dependencies.

src/jsonSchema.ts
TheDSCPL: We do. `typeof null === 'object'`. I've made that mistake once or twice before :)

TheDSCPL:
There. Build successful and all of the reviews were committed as well

src/jsonSchema.ts
TheDSCPL: I assumed `validateAllFieldCriteria` is set true when `criteriaMode==='all'` and false otherwise (i.e. `firstError`). Already committed the change.

src/jsonSchema.ts
TheDSCPL: Removed the first `replace` call. Change already committed.

bluebill1049:
That's awesome! Thanks very much, @TheDSCPL i will get a beta version for you all first during the week.

src/jsonSchema.ts
bluebill1049: I would mostly like to remove those type export in the final build as consistency with other resolvers, you can leave that to me tho.

TheDSCPL: Uhm ok. `json-schema` is just a type definition package. I only wanted to add some form of type check to the schema like all the other implementations. The difference is that the schema is implied by the other libraries but ajv takes a schema with `any` type. It doesn't define the type of the schema. Instead it just throws if the schema is invalid. That's why I added the type.

bluebill1049:
here is the beta version: `@hookform/resolvers@0.2.0-beta.1`, please help with the testing 👍

jorisre:
@bluebill1049  or @TheDSCPL is the beta version working for you ? 
I get the following error 
> Could not find module in path: '@hookform/resolvers/dist/compile' relative to '/node_modules/@hookform/resolvers/dist/index.esm.js'

When I try with ajv or yup, here a codesandbox : https://codesandbox.io/s/rhf-resolvers-beta-test-t550f
No error with 0.1.1

bluebill1049:
> @bluebill1049 or @TheDSCPL is the beta version working for you ?
> I got the following error
> 
> > Could not find module in path: '@hookform/resolvers/dist/compile' relative to '/node_modules/@hookform/resolvers/dist/index.esm.js'
> 
> When I try with ajv or yup, here a codesandbox : https://codesandbox.io/s/rhf-resolvers-beta-test-t550f
> No error with 0.1.1

hmmm, the only change i can think of its the package.json file.

TheDSCPL:
@bluebill1049 It's strange because there wasn't any change made to yup's library as you can see in the [Files changed](https://github.com/react-hook-form/resolvers/pull/50/files) tab. It's most likely a problem with module resolution in the esm bundle. We'll have to check the bundler's configs but it can't be me because I'm not familiar with rollup and I can't study the documentation atm

bluebill1049:
> @bluebill1049 It's strange because there wasn't any change made to yup's library as you can see in the [Files changed](https://github.com/react-hook-form/resolvers/pull/50/files) tab. It's most likely a problem with module resolution in the esm bundle. We'll have to check the bundler's configs but it can't be me because I'm not familiar with rollup and I can't study the documentation atm

yes, confirmed it's not related to that PR.

TheDSCPL:
I believe I know what it is. Remember when I was saying that the flag `esModuleInterop` should be set to true on the CI build's `tsconfig.json`? This is that biting us in the butt.

TheDSCPL:
That's why this error didn't happen on my PC. I set that option to true and it worked. I had to later set it to false for the CI build to be happy but it's not correct.

TheDSCPL:
Can you change the `tsconfig.json` file used by the CI build?

TheDSCPL:
Btw, when you do change it, please tell me so I can make another PR and change ajv's import to `import ajv from 'ajv'` instead of `import * as ajv from 'ajv'`

bluebill1049:
https://github.com/react-hook-form/resolvers/pull/58 WIP for the fix

bluebill1049:
> @bluebill1049 or @TheDSCPL is the beta version working for you ?
> I got the following error
> 
> > Could not find module in path: '@hookform/resolvers/dist/compile' relative to '/node_modules/@hookform/resolvers/dist/index.esm.js'
> 
> When I try with ajv or yup, here a codesandbox : https://codesandbox.io/s/rhf-resolvers-beta-test-t550f
> No error with 0.1.1

Hey @jorisre if you working on fixing the build, I would recommend skipping this merge commit. I haven't got any luck so far for a fix.

gwendall:
Don't know if that's related, but I get the following error:

```
Can't resolve './cache' in '/Users/gwendall/Code/biogen/mybiogen/backoffice/node_modules/@hookform/resolvers/dist'
```

bluebill1049:
> Don't know if that's related, but I get the following error:
> 
> ```
> Can't resolve './cache' in 'node_modules/@hookform/resolvers/dist'
> ```

don't think so, this PR is not published just yet.

TheDSCPL:
@bluebill1049 Is it possible for you to reopen this PR so I can push another change or do I need to make another PR?

bluebill1049:
> @bluebill1049 Is it possible for you to reopen this PR so I can push another change or do I need to make another PR?

hey @TheDSCPL please send another PR, i can't reopen here. thank very much for helping out mate.

PR #57, opened 2020-09-15
Feature/zod resolver

jorisre:
Hey 👋🏻 

Little PR to add [Zod](https://github.com/vriad/zod) resolver.

Let me know if need some changes !


julienfouilhe:
What??? I was about to submit my PR for zod support, and I see yours, submitted the same day! What a coincidence!

src/zod.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  return zodError.errors.reduce<Record<string, any>>(
+    (previous, { path, message, code: type }) => {
+      const currentPath = convertArrayToPathName(path);
+
+      previous[currentPath] = {
+        message,
+        type,
+      };
```
julienfouilhe, line 28: mutating the accumulator is usually not a good practice, this makes the reducer not pure.

src/zod.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    {},
+  );
+};
+
+export const zodResolver = <TFieldValues extends FieldValues>(
+  schema: Schema<any>,
+  options?: ParseParams,
+): Resolver<TFieldValues> => async (
```
julienfouilhe, line 64: Otherwise type will not be infered from the Schema?

```suggestion
export const zodResolver = <TFieldValues extends FieldValues>(
  schema: Schema<TFieldValues>,
  options?: ParseParams,
): Resolver<TFieldValues> => async (
```

src/zod.ts
```diff
@@ -0,0 +1,81 @@
...
+} from 'react-hook-form';
+import { Schema, ZodError } from 'zod';
+import { ParseParams } from 'zod/lib/src/parser';
+import convertArrayToPathName from './utils/convertArrayToPathName';
+
+const parseErrorSchema = (
+  zodError: ZodError,
+  validateAllFieldCriteria: boolean,
+) => {
+  if (!Array.isArray(zodError.errors) || zodError.isEmpty) {
```
julienfouilhe, line 17:
```suggestion
  if (!zodError.isEmpty) {
```

I think we can trust zod typings to be true. errors is always an array

src/zod.ts
julienfouilhe:
```suggestion
export const zodResolver = <TFieldValues extends FieldValues, ZodTypeDef extends ZodTypeDef>(
  schema: Schema<TFieldValues, ZodTypeDef>,
  options?: ParseParams,
): Resolver<TFieldValues> => async (
```

actually would even be better, otherwise we couldn't pass a schema with a different type def

jorisre: Thanks for reviewing, it looks better :)

bluebill1049:
oh wow! thanks guys, this is awesome. I just started the project yesterday and now there is a resolver for hook form. amazing.

I will sort out the AJV one first before looking into this one, give me sometimes, but I am sure between @julienfouilhe & yourself, this will be read by the time i am looking at it.

src/zod.ts
vriad: you can! :)

src/zod.ts
```diff
@@ -0,0 +1,79 @@
...
+    },
+    {},
+  );
+};
+
+export const zodResolver = <
+  TFieldValues extends FieldValues,
+  TZodTypeDef extends ZodTypeDef
+>(
+  schema: Schema<TFieldValues, TZodTypeDef>,
```
vriad, line 60: I recommend writing generic functions that accept Zod schemas like so:

```ts
const zodResolver = <T extends z.ZodSchema<any,any>>(
  valueSchema: T,
): Resolver<Infer<T>> => { // you'll have to import `Infer` from 'zod'
  // do stuff
}
```

I talk about the reasons a bit here: https://github.com/vriad/zod/issues/153#issuecomment-692903583

jorisre: Thanks for your recommandations !
I've used `TypeOf` from zod instead of `Infer` to avoid this import:  `import { Infer } from 'zod/lib/src/types/base';`
I'm wrong ?

vriad: Oops typo: it should be `infer` (lower case)

And they're exactly the same 😅   I added `infer` as an alias because I thought it was a little more intuitive.

Screen Shot 2020-09-16 at 12 02 46 PM

jorisre: Thanks ! updated ! :)

jorisre:
Updated !  ✅

README.md
```diff
@@ -133,6 +133,34 @@ const App = () => {
 };
 ```
 
+### [Zod](https://github.com/vriad/zod)
```
bluebill1049, line 136: shall we move this on top of Joi please (or even superstruct)?

bluebill1049 (approved):
Looks good to me.

README.md
jorisre: Done ✅

bluebill1049: thank you

bluebill1049:
v1 is going to have zod resolver :)

bluebill1049:
Thanks very much for your hard work @jorisre and everyone else who helped with this PR.

PR #58, opened 2020-09-18
fix ajv

bluebill1049:


jorisre:
Thanks @bluebill1049 ! This will fix umd bundle :)

bluebill1049:
> Thanks @bluebill1049 ! This will fix umd bundle :)

doesn't looks like it's working... :(

bluebill1049:
Screen Shot 2020-09-18 at 4 15 03 pm

I think the issue is import Ajv, i will remove the import and should should construct the schema.

bluebill1049:
@TheDSCPL FYI

src/jsonSchema.ts
```diff
@@ -1,8 +1,8 @@
 import * as Ajv from 'ajv';
 import { appendErrors, Resolver, transformToNestObject } from 'react-hook-form';
-import type { ErrorObject, Options, ValidateFunction } from 'ajv';
-import type { JSONSchema4, JSONSchema6, JSONSchema7 } from 'json-schema';
-import type {
+import { ErrorObject, Options, ValidateFunction } from 'ajv';
```
TheDSCPL, line 3: Why did you delete the "type"? It was correct

src/jsonSchema.ts
```diff
@@ -1,8 +1,8 @@
 import * as Ajv from 'ajv';
```
TheDSCPL, line 1: When you add `esModuleInterop: true` to the `tsconfig.json`, `* as ` should be removed and all the type errors will disapear.

src/jsonSchema.ts
```diff
@@ -79,7 +79,8 @@ export const jsonSchemaResolver = <TFieldValues extends Record<string, any>>(
     // compile only if necessary. otherwise return the memoized validation function
     if (!validate || allErrors !== validateAllFieldCriteria) {
       allErrors = validateAllFieldCriteria;
-      validate = new Ajv({
+      // @ts-ignore
+      validate = new Ajv.default({
```
TheDSCPL, line 83: This shouldn't be done. TS was made by very clever people who know how the modules systems work. If it's an error, it won't work on runtime. In this case changing the import and changing this line back to how it was will remove the error.

src/jsonSchema.ts
```diff
@@ -88,6 +89,7 @@ export const jsonSchemaResolver = <TFieldValues extends Record<string, any>>(
       }).compile(validationSchema);
     }
 
+    // @ts-ignore
```
TheDSCPL, line 92: What was the error here? Btw for future reference using `@ts-expect-error` is recommended instead of `@ts-ignore`. But even that should seldom be used.

src/jsonSchema.ts
bluebill1049: Screen Shot 2020-09-18 at 4 25 07 pm

src/jsonSchema.ts
bluebill1049: Screen Shot 2020-09-18 at 4 25 32 pm

src/jsonSchema.ts
bluebill1049: i build it and failed without it.

src/jsonSchema.ts
bluebill1049: at this stage, I am trying to get it working first.

bluebill1049:
@TheDSCPL I am going to try to pull Ajv out.
Screen Shot 2020-09-18 at 4 26 26 pm

TheDSCPL:
What do you mean? Delete it?

bluebill1049:
> What do you mean? Delete it?

no, move the new Ajv out from the resolver, so we don't need to import Ajv. that's the root issue.

bluebill1049:
> What do you mean? Delete it?

unless you have a better strategy to fix it. as it's probably a blocker right now.

TheDSCPL:
> no, move the new Ajv out from the resolver, so we don't need to import Ajv. that's the root issue.

I think that will have to the next step. As I've said before, I think separating the resolver implementations into different sub-packages under `@hookform` will be the only effective solution for the users to not install 3 extra packages (and their respective dependencies, dependencies of dependencies, ...) for just using one of the implementations.

Anyway, that is not the issue here, though.

jorisre:
No sure about this solution but to make it works, you can add `external: ['ajv']` to rollup config : 

```js
  const config = {
    input: options.input,
    external: ['ajv'], // <= It works !
    output: {
      file: outputName,
      format: options.format,
      name: umdName,
      sourcemap: true,
      globals: {
        'react-hook-form': 'ReactHookForm',
      },
      exports: 'named',
    },
...
```

TheDSCPL:
@bluebill1049 [There you go](https://github.com/TheDSCPL/resolvers/actions/runs/260663039)

TheDSCPL:
@jorisre if you set ajv as an external it won't be bundled and hence not work on the browser

bluebill1049:
hey @TheDSCPL you happy to pull that commit out from master? I will create a PR again from that commit until we find a fix for it.

TheDSCPL:
I'll make a new PR with the new changes [that passed the CI build](https://github.com/TheDSCPL/resolvers/actions/runs/260663039).

Issue #87, opened 2020-12-08
File name error ./supertruct.d.ts


oscarlgz:
[This file](https://github.com/react-hook-form/resolvers/blob/master/supertruct.d.ts) needs to be renamed to `superstruct.d.ts`.

Thanks for your work!

jorisre:
Good catch ! Fixed in https://github.com/react-hook-form/resolvers/pull/89

PR #89, opened 2020-12-08
fix: superstruct filename typo

jorisre:
Related to https://github.com/react-hook-form/resolvers/issues/87


bluebill1049:
Thank you @jorisre

PR #90, opened 2020-12-10
Fixing and standardizing Zod example

colinhacks:
For some reason the Zod example implemented a different form from the other 3 examples, so I decided to standardize it wiht the others. In the process of doing that I realized the sample was buggy because the Zod schema didn't convert the `age` string into a number. This makes sense because Zod v1 (which the author of the resolver was using) didn't support value transformations. I was planning to upgrade the example to use Zod 2 (which is capable of converting the string to a number) but then I saw that support for `valueAsNumber` was just introduced! Which is SUCH a gamechanger, thanks Bill! Because of that I was able to fix this example without dropping support for Zod v1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,29 +74,36 @@ TypeScript-first schema validation with static type inference
 
 [![npm](https://img.shields.io/bundlephobia/minzip/zod?style=for-the-badge)](https://bundlephobia.com/result?p=zod)
 
-```typescript jsx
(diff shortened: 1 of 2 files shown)


colinhacks:
I also changed `tsConfig` to `tsconfig` to get rid of an warning message that appeared when you run `npm run test`.

Screen Shot 2020-12-10 at 2 49 26 PM

bluebill1049:
sweet, thanks @colinhacks

Issue #91, opened 2020-12-11
Cannot resolve symbol 'yupResolver' 


LuisFilipePedroso:
Hey, I'm trying to import yupResolver in my project, where I'm using Typescript but I'm getting this error:
**Cannot resolve symbol 'yupResolver'** 
 
I'm importing yupResolver like this:
`import { yupResolver } from '@hookform/resolvers/yup';`

Is there something that I'm doing wrong?

Version of @hookform/resolvers: 1.1.2
Version of react-hook-form: 6.12.2
Version of Yup: 0.32.8

jorisre:
👋🏻  @LuisFilipePedroso 

I think, it's related to https://github.com/react-hook-form/resolvers/issues/86. I've started working on it yesterday.


Could you confirm it's the same error ? Or provide a reproduction example in codesandbox please ?

LuisFilipePedroso:
Hey @jorisre , I saw this issue but I guess that he was able to import yupResolver, in my case, I'm not being able to import, what is strange. 

You can check the code here:
https://github.com/LuisFilipePedroso/Youtube-React-Form

Form is here:
https://github.com/LuisFilipePedroso/Youtube-React-Form/blob/master/src/components/Form/index.tsx

To run, just install the dependencies and execute the commands:

npm run api 
npm run start

jorisre:
Thanks ! I'll look at it later :)

LuisFilipePedroso:
Thanks.

jorisre:
This issue is related to #86. Yup versions are different that's why this error occur.

jorisre:
@LuisFilipePedroso 
Your issue should be fixed
Yup ^0.32.0 is now supported ! [1.3.0](https://github.com/react-hook-form/resolvers/releases/tag/v1.3.0)

Issue #86, opened 2020-12-08
Type error in yup resolver


chemaxa:
Hi everyone. Thank you for a great tool. 
I tried to use yup resolver with typescript. But i faced with typescript compilation error.

```sh
i ｢wdm｣: Compiled with warnings.
ERROR in D:/Repositories/form/node_modules/@hookform/resolvers/dist/yup.d.ts(3,86):
TS2707: Generic type 'ObjectSchema<TShape, TContext, TIn, TOut>' requires between 1 and 4 type arguments.

ERROR in D:/Repositories/form/node_modules/@hookform/resolvers/dist/yup.d.ts(3,109):
TS2694: Namespace '"D:/Repositories/form/node_modules/yup/lib/index"' has no exported member 'Lazy'.

ERROR in D:/Repositories/form/node_modules/@hookform/resolvers/dist/yup.d.ts(3,134):
TS2694: Namespace '"D:/Repositories/form/node_modules/yup/lib/index"' has no exported member 'ValidateOptions'.
```

My packages:
"yup": "^0.32.5"
"react-hook-form": "^6.12.2",  
"@hookform/resolvers": "^1.1.1",

I tried to use @types/yup but it doesnt help to me
"@types/yup": "^0.29.10",

Can anybody to help me? 
How i can to fix this issue?

jorisre:
Hi @chemaxa 👋🏻 

Could you please provide a reproduction in a [codesandbox](https://codesandbox.io/) ? Thanks

chemaxa:
Hi, i cant reproduce it on codesandbox. But i prepared repository with code sample.

https://github.com/chemaxa/hookresolvers-ts-bug

For reproduce just 

npm install 

npm run start:standalone

I used 
https://single-spa.js.org/docs/create-single-spa
For creating demo project with react, typescript settings.

Also i tried to set esModuleInterop to false, and this bug is out, but this is not applicable for my workflow. 

I dont understand how you used yup objectshema type, when this type is generic and required 1 or 4 params. 
https://github.com/jquense/yup/blob/01c69305424db349ff701fae7cdb31e569c38b6e/src/object.ts#L75

chemaxa:
I reproduced it in codesandbox.

I just imported yup and tried use types as in yuor yup.d.ts file and got same error

https://codesandbox.io/s/yup-typescript-error-d1n3j?file=/src/index.ts

chemaxa:
I think i understand what the reason of this bug, you are used  
[...]

PR #92, opened 2020-12-13
Support Yup ^0.32.0

jorisre:
Related to https://github.com/react-hook-form/resolvers/issues/86 & https://github.com/react-hook-form/resolvers/issues/91

[Yup@0.32.0](https://github.com/jquense/yup/blob/master/CHANGELOG.md#0320-2020-12-03) add Typescript support. `@types/yup` isn't required anymore.


src/yup.ts
```diff
@@ -1,13 +1,21 @@
-import { Resolver, transformToNestObject, FieldValues } from 'react-hook-form';
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+import { Resolver, transformToNestObject } from 'react-hook-form';
 import Yup from 'yup';
 
+/**
+ * From 0.32.0, Yup add TypeScript support and `path` typing is optional that's why we have `@ts-expect-error`
```
jorisre, line 6: @bluebill1049 I'm not satisfied, if you've a better solution

bluebill1049: > FYI: `path`: a string, indicating where there error was thrown. `path` is empty at the root level.

what do you mean by that?

bluebill1049: Sorry, I haven't played with the latest Yup, probably lack of context.

jorisre: This message is from https://github.com/jquense/yup#validationerrorerrors-string--arraystring-value-any-path-string

bluebill1049: oh, `path` is empty, basically means the object key is actually the path?

jorisre: Yes, the new type for ValidationError is:

```ts
export default class ValidationError extends Error {
    value: any;
    path?: string;
    type?: string;
    errors: string[];
    params?: Params;
    inner: ValidationError[];
    static formatError(message: string | ((params: Params) => string) | unknown, params: Params): any;
    static isError(err: any): err is ValidationError;
    constructor(errorOrErrors: string | ValidationError | ValidationError[], value?: any, field?: string, type?: string);
}
```


As you can see, `path` & `type` could be `undefined`

bluebill1049: Hmm, interesting, let's go with the `ignore` for now 👍 it's a bit inconsistent for the rest of the schema libs.

jorisre: Ok, PR is ready to review :)

src/yup.test.ts
```diff
@@ -42,7 +42,7 @@ const errors = {
   ],
 };
 
-const schema = yup.object().shape({
+const schema = yup.object({
```
bluebill1049, line 45: oh, shape is gone. 👍

src/yup.test.ts
```diff
@@ -119,26 +120,33 @@ describe('yupResolver', () => {
         createdOn: null,
         foo: [{ loose: null }],
       };
-      const resolve = await yupResolver(schema)(data, {}, true);
-      expect(resolve).toMatchSnapshot();
-      expect(resolve.errors['foo'][0]['loose']).toBeDefined();
-      expect(resolve.errors['foo'][0]['loose'].types).toMatchInlineSnapshot(`
+
+      const output = await yupResolver(schema)(
+        // @ts-expect-error
```
bluebill1049, line 125: minor (don't have to fix it): Is this due to reuse the same schema? maybe worth to use a new schema instead?

src/yup.test.ts
```diff
@@ -233,6 +244,8 @@ describe('validateWithSchema', () => {
           return min ? schema.min(6) : schema;
         }),
     });
+
+    // @ts-expect-error
```
bluebill1049, line 248: assume this is the same issue above?

bluebill1049 (approved):
Awesome stuff! 🥳 Thank you very much @jorisre

src/yup.test.ts
jorisre: This is normal because `yupResolver` have a [better type support now :) ](https://github.com/react-hook-form/resolvers/pull/92/files/8701c7062718187ce7bbe1693a2d1a4105f216c9#diff-c1efacb5b2a76e9e4b9c1fd06e62bcac926439ffc153d798f5e440e79f3a2882)

The schema is good but passed data are intentionally false, that's why TS throw an error.

src/yup.ts
```diff
@@ -38,41 +53,44 @@ const parseErrorSchema = (
...
+
+type ValidateOptions<T extends Yup.AnyObjectSchema> = Parameters<
+  T['validate']
+>[1];
 
-export const yupResolver = <TFieldValues extends FieldValues>(
-  schema: Yup.ObjectSchema | Yup.Lazy,
-  options: Omit<Yup.ValidateOptions, 'context'> = {
+export const yupResolver = <T extends Yup.AnyObjectSchema>(
+  schema: T,
```
sukjae, line 66: hi! 

i would like to hear about the reason why `Yup.Lazy` type schema is removed. 
from my understanding, from #31 we begun to support Lazy type schema but was removed in this PR. 

if it was meant to be, i would love hear about it. 
otherwise, i would make PR to fix it!

thanks~ @jorisre

jorisre: @sukjae 👋🏻 
It's a mistake

Feel free to send a PR ;)

sukjae: hi @jorisre 👋
nice! i made PR :) 
thanks~
#135

Issue #71, opened 2020-11-05
Uncaught ReferenceError: exports is not defined


batamire:
trying to:
```
import { zodResolver } from "@hookform/resolvers/zod";
```

gives me:

```
Uncaught ReferenceError: exports is not defined
    at Module../node_modules/@hookform/resolvers/dist/zod.js (zod.js:11)
    at __webpack_require__ (bootstrap:789)
    at fn (bootstrap:100)
    at Object../node_modules/@hookform/resolvers/zod.js (zod.js:1)
    at __webpack_require__ (bootstrap:789)
    at fn (bootstrap:100)
    ...
```

This is a Rails webpacker + typescript app.

bluebill1049:
more detail... which version and etc... can you reproduce that in CSB?

batamire:
`@hookform/resolvers": "^1.0.0"`

I'm afraid I cannot reproduce it as this is a large Rails app with TypeScript React and GraphQL. Could it be the way you export modules? I really lack JS packaging knowledge...

my babel.config.js:
```
module.exports = function(api) {
  var validEnv = ["development", "test", "production"];
  var currentEnv = api.env();
  var isDevelopmentEnv = api.env("development");
  var isProductionEnv = api.env("production");
  var isTestEnv = api.env("test");

  if (!validEnv.includes(currentEnv)) {
    throw new Error(
      "Please specify a valid `NODE_ENV` or " +
        '`BABEL_ENV` environment variables. Valid values are "development", ' +
        '"test", and "production". Instead, received: ' +
        JSON.stringify(currentEnv) +
        "."
    );
  }

  return {
    presets: [
      isTestEnv && [
        "@babel/preset-env",
        {
          targets: {
            node: "current"
          },
          modules: "commonjs"
        },
        "@babel/preset-react"
      ],
      (isProductionEnv || isDevelopmentEnv) && [
        "@babel/preset-env",
        {
          forceAllTransforms: true,
          useBuiltIns: "entry",
          corejs: 3,
          modules: false,
          exclude: ["transform-typeof-symbol"]
        }
      ],
      [
        "@babel/preset-react",
        {
          development: isDevelopmentEnv || isTestEnv,
          useBuiltIns: true
        }
      ],
      ['@babel/preset-typescript', { 'allExtensions': true, 'isTSX': true }]
    ].filter(Boolean),
    plugins: [
      "babel-plugin-macros",
      "@babel/plugin-syntax-dynamic-import",
[...]

PR #95, opened 2020-12-29
esm support to fix bundler incompatibility

benallfree:
This package `dist` is currently not compatible with bundlers.

`tsc` outputs a commonjs format, but that is for node environments, not browser. It causes `Object.defineProperty(exports, "__esModule", { value: true });` to be emitted, which causes an error in a browser.

I updated `tsconfig.json` to output `esnext` instead, and that fixes the problem.

See for more info: https://blog.rendall.dev/posts/2019/1/14/problem-typescript-adds-objectdefinepropertyexports-esmodule-value-true/


jorisre:
👋🏻  @benallfree 
Thanks for your work!
Do you also have this [problem](https://github.com/react-hook-form/resolvers/issues/71) ?

benallfree:
@jorisre Yes, this fixes #71. Thanks for catching that!

jorisre (approved):
LGTM Thanks @benallfree !

bluebill1049:
Thanks @jorisre shall we do a release with it today?

jorisre:
> Thanks @jorisre shall we do a release with it today?

Yes you can :)

bluebill1049:
@jorisre sent you a private message at twitter.

benallfree:
Any thought on when the updated npm package will be published? I have to maintain my own branch with a `build` folder until the package is published

Issue #100, opened 2020-12-31
Build error after upgrade to 1.3.1 in Next.js


nghieptiki:
Hi guys,

I have an SSR app working on Node environments, after upgrade to 1.3.1 and build error:

>[2020-12-31T02:56:35.428Z] #16 90.33 > Build error occurred
>[2020-12-31T02:56:35.428Z] #16 90.34 /app/node_modules/@hookform/resolvers/dist/yup.js:2
>[2020-12-31T02:56:35.428Z] #16 90.34 import { transformToNestObject, } from 'react-hook-form';
>[2020-12-31T02:56:35.428Z] #16 90.34 ^^^^^^
>[2020-12-31T02:56:35.428Z] #16 90.34 
>[2020-12-31T02:56:35.428Z] #16 90.34 SyntaxError: Cannot use import statement outside a module
>[2020-12-31T02:56:35.428Z] #16 90.34     at wrapSafe (node:internal/modules/cjs/loader:1024:16)
>[2020-12-31T02:56:35.428Z] #16 90.34     at Module._compile (node:internal/modules/cjs/loader:1072:27)

I think the problem related to https://github.com/react-hook-form/resolvers/pull/95 PR.

joshmanders:
Getting the same thing. Downgrading to `1.2.0` fixes it.

jorisre:
Thanks for reporting.
I'm working on resolvers's V2 which will have a better module supports.

As a workaround,  you can use 1.3.0

Rykuno:
> Thanks for reporting.
> I'm working on resolvers's V2 which will have a better module supports.
> 
> As a workaround, you can use 1.3.0

Bumping down to 1.3.0 does not solve for me. Thanks for your hard work, looking forward to V2 with module support. While this is currently putting a hold on validation we have been overall loving react-hook-form. 
```
{
   "@hookform/resolvers": "^1.3.0",
    "react-hook-form": "^6.14.0",
    "next": "^10.0.4",
}
```

msheakoski:
@Rykuno You need to use `"1.3.0"`, not `"^1.3.0"`. Using the ^ means greater than or equal to 1.3.0 but less than 2.0.0. You can read more about it here: https://docs.npmjs.com/cli/v7/using-npm/semver#caret-ranges-123-025-004

Rykuno:
[...]

PR #104, opened 2021-01-02
V2 - better modules support

jorisre:
First of all, huge thank to the [preact](https://github.com/preactjs/preact) hard work. This PR is heavily inspired of their modules structure.

Good news, no breaking changes ! users can import resolvers as usual 🎉 

- move resolvers into separate folders
- add exports map in package.json + support node 13 _(thank to preact)_
- build: use microbundle + severals updates
- test: update config
- remove IE11
- use semantic-release default config

Tested with Next.js and CRA with success ✅  ! Others formats will be tested after first beta release

Solves https://github.com/react-hook-form/resolvers/issues/71 & https://github.com/react-hook-form/resolvers/issues/100


package.json
```diff
@@ -88,19 +136,14 @@
   "peerDependencies": {
     "react-hook-form": ">=6.6.0"
   },
-  "release": {
```
jorisre, line 91: By removing this, this allows us to use semantic-release's default config, so master, beta branches will be automatically published.

In our case, beta will create 2.0.0-beta.X version

https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#pre-release-branches

bluebill1049:
This is super exciting!

jorisre:
@bluebill1049 PR ready for review 🙂

bluebill1049:
Screen Shot 2021-01-04 at 8 40 41 pm

Wow, this is cool!

joi/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "React Hook Form validation resolver: joi",
+  "main": "dist/joi.js",
+  "module": "dist/joi.module.js",
+  "umd:main": "dist/joi.umd.js",
+  "source": "src/index.ts",
+  "types": "dist/index.d.ts",
+  "license": "MIT",
+  "peerDependencies": {
+    "react-hook-form": ">=6.6.0",
+    "@hookform/resolvers": "^1.0.0"
```
bluebill1049, line 15: dependent itself?

package.json
```diff
@@ -1,36 +1,89 @@
...
+      "umd": "./vest/dist/vest.umd.js",
+      "import": "./vest/dist/vest.mjs",
+      "require": "./vest/dist/vest.js"
+    },
+    "./superstruct": {
+      "browser": "./superstruct/dist/superstruct.module.js",
+      "umd": "./superstruct/dist/superstruct.umd.js",
+      "import": "./superstruct/dist/superstruct.mjs",
+      "require": "./superstruct/dist/superstruct.js"
+    },
```
bluebill1049, line 47: nice!

package.json
```diff
@@ -1,36 +1,89 @@
...
-    "build": "tsc && rollup -c",
-    "lint": "eslint '**/*.{js,ts}'",
-    "lint:fix": "yarn lint -- --fix",
-    "lint:types": "tsc --noEmit",
-    "test": "jest --runInBand",
-    "test:watch": "yarn test --watchAll --coverage"
+    "prepare": "run-s build && check-export-map",
+    "build": "npm-run-all --parallel build:*",
+    "build:src": "microbundle build",
+    "build:zod": "microbundle --cwd zod --globals '@hookform/resolvers=hookformResolvers'",
```
bluebill1049, line 76: That's new to me, `--globals`

bluebill1049 (approved):
This is too good! Does this need a breaking change? if not maybe we can keep it under v1, until v2 with breaking changes (those new attributes, arguments, and functionalityies)?

joi/package.json
jorisre: Yes, to import `convertArrayToPathName`, inspiration from [preact](https://github.com/preactjs/preact/blob/master/hooks/package.json#L21)

package.json
jorisre: It specify variable name for external imports in umd/iife bundles

jorisre:
> This is too good! Does this need a breaking change? if not maybe we can keep it under v1, until v2 with breaking changes (those new attributes, arguments, and functionalityies)?

I suggest to publish a V2 beta because there are lot of changes (project structure, bundles etc...) And we can advise users to try beta and collect feedbacks. What do you think ?
I’m thinking about a powerful v2 without any breaking changes for developers :) 🤞🏻

bluebill1049:
> > This is too good! Does this need a breaking change? if not maybe we can keep it under v1, until v2 with breaking changes (those new attributes, arguments, and functionalityies)?
> 
> I suggest to publish a V2 beta because there are lot of changes (project structure, bundles etc...) And we can advise users to try beta and collect feedbacks. What do you think ?
> I’m thinking about a powerful v2 without any breaking changes for developers :) 🤞🏻

sounds good to me! The only thing to beware that their argument of resolver may change with v7, we didn't document the third argument in the readme tho, just in case someone is using it.

joi/package.json
bluebill1049: oh cool! 👍

Issue #97, opened 2020-12-30
Type problems after update resolver to 1.3.0


william-hotmart:
After update @hookform/resolvers to the version 1.3.0, this code above no longer works.

The TS compiler points out the error:

```
Type 'Resolver<AssertsShape<Assign<Record<string, AnySchema<any, any, any> | Reference<unknown> | Lazy<any, any>>, { title: RequiredStringSchema<string, Record<string, any>>; buttonTitle: RequiredStringSchema<...>; exibitionMode: RequiredStringSchema<...>; link: RequiredStringSchema<...>; duration: NumberSchema<...>; }>>,...' is not assignable to type 'Resolver<FormData, object>'.
  Types of parameters 'values' and 'values' are incompatible.
    Type 'FormData' is not assignable to type '{ [x: string]: never; title: string; buttonTitle: string; exibitionMode: string; link: string; duration: number; }'.
      Index signature is missing in type 'FormData'.ts(2322)
form.d.ts(39, 5): The expected type comes from property 'resolver' which is declared here on type 'Partial<{ mode: "onBlur" | "onChange" | "onSubmit" | "onTouched"
```

If I comment the lines of _exibitionMode_ and _colorBackground_ the error disappears.

This is the code that I'm working with

```js
import { useForm } from 'react-hook-form'
import { yupResolver } from '@hookform/resolvers/yup'
import { object, string, number } from 'yup'

enum Color {
  DARK = 'dark',
  LIGHT = 'light',
  PURPLE = 'purple',
  ORANGE = 'orange',
  RED = 'red',
  GREEN = 'green'
}

enum ExibitionMode {
  INFINITY = 'infinity',
  LIMITED = 'limited'
}

interface FormData {
  title: string,
  buttonTitle: string,
  exibitionMode: ExibitionMode,
  link: string,
  duration: number,
  colorBackground: Color
}

const TITLE_MAX_LENGTH = 96
const BUTTON_TITLE_MAX_LENGTH = 28

const schema = object().shape({
  title: string().max(TITLE_MAX_LENGTH).required(),
  buttonTitle: string().max(BUTTON_TITLE_MAX_LENGTH).required(),
  exibitionMode: string().required(),
  link: string().required().url(),
  duration: number().when('exibitionMode', (exibitionMode, schema) => {
    return exibitionMode === ExibitionMode.LIMITED ? schema.min(1).required() : schema
  })
})

const Test = () => {
  const { register, handleSubmit } = useForm<FormData>({
[...]

PR #106, opened 2021-01-04
fix: yup resolver TypeScript errors

jorisre:
Related to https://github.com/react-hook-form/resolvers/issues/97

I just put the original types back for the yup resolver.
I plan to improve them in V2 for all resolvers


bluebill1049 (approved):
👍

Issue #88, opened 2020-12-08
Feature request: support synchronous validation


Smona:
We are using yup along with react-hook-form, and almost all of our schemata support synchronous validation. Unfortunately, there's no way to use `schema.validateSync` instead of `schema.validate` through the resolver.

While the performance impact of this is negligible, inserting an event loop between form submission and `handleSubmit` being called adds an unnecessary layer of complexity to our tests.

It would be great if we could pass an option to validate the schema synchronously. Better yet, if the resolver attempted to validate synchronously and fell back on asynchronous validation if required.

bluebill1049:
`resolver: async () => schema.validateSync(data)`

have you tried to build your own?

Smona:
No, not yet. I could, but then I would have to copy-paste the source code for the error parser and other imports of the resolver. Which kind of defeats the purpose of the repo :sweat_smile:

Smona:
I suppose I'm assuming that `useForm` would support a synchronous validation resolver. If that's not the case, I can see why this would be harder to implement.

jorisre:
Hi @Smona 👋🏻

FIY Yup sync validation added in the [v2.0.0-beta.5](https://github.com/react-hook-form/resolvers/releases/tag/v2.0.0-beta.5)
You can try it by installing `yarn add @hookform/resolvers@v2.0.0-beta.5`

GollyJer:
HI @jorisre 👋

Stumbled on this today looking for a synchronous solution. 
I have 
```
"@hookform/resolvers": "2.0.0-beta.5",
"react-hook-form": "^6.14.2",
```

And the following code.
```javascript 
const { control, errors, watch, formState } = useForm<FormInputs>({
    mode: 'onChange',
    reValidateMode: 'onChange',
    defaultValues,
    resolver: yupResolver(
      Yup.object().shape({
        syringeUnits: Yup.number().required().moreThan(0),
        syringeTickMarks: Yup.number().required().moreThan(0),
        powderMilligrams: Yup.number().required().moreThan(0),
        waterMilliliters: Yup.number().required().moreThan(0),
        desiredDoseAmount: Yup.number().required().moreThan(0),
      }),
      undefined,
      { mode: 'sync' }
    ),
  });

  const formValues = watch();
  console.log(formValues);
```

I'm still seeing formValues print the old and new value when typing in a field. 
Am I doing something wrong?

Thanks for the work on this! It's going to be super helpful.

bluebill1049:
[...]

PR #114, opened 2021-01-20
Zod sync validation

jorisre:
Related to https://github.com/react-hook-form/resolvers/issues/88

Starting with Zod, same for others resolvers coming :)

### DIFF
--- a/zod/src/types.ts
+++ b/zod/src/types.ts
@@ -8,7 +8,8 @@ import type { ParseParams } from 'zod/lib/src/parser';
 
 export type Resolver = <T extends z.ZodSchema<any, any>>(
   schema: T,
-  options?: ParseParams,
+  schemaOptions?: ParseParams,
+  resolverOptions?: { mode: 'async' | 'sync' },
 ) => <TFieldValues extends FieldValues, TContext>(
   values: UnpackNestedValue<TFieldValues>,
   context?: TContext,
--- a/zod/src/__tests__/__snapshots__/zod.ts.snap
+++ b/zod/src/__tests__/__snapshots__/zod.ts.snap
@@ -40,6 +40,46 @@ Object {
 }
 `;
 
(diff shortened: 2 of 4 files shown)


zod/src/types.ts
```diff
@@ -8,7 +8,8 @@ import type { ParseParams } from 'zod/lib/src/parser';
 
 export type Resolver = <T extends z.ZodSchema<any, any>>(
   schema: T,
-  options?: ParseParams,
+  schemaOptions?: ParseParams,
+  resolverOptions?: { mode: 'async' | 'sync' },
```
bluebill1049, line 12: to use this feature, i guess users will have to this:

`resolver: async (data, options) => await zodResolver(schema, options, 'async')`

jorisre: No they have to this:
```ts
useForm({
  resolver: zodResolver(schema, options, {mode: 'sync'})
})
```

 Or if you prefer, we can do the following:

```ts
interface Options {
  schema?: ParseParams;
  resolver?: { mode: 'async' | 'sync' }
}

export type Resolver = <T extends z.ZodSchema<any, any>>(
  schema: T,
  options?: Options
) => { ... }
```

So user will have to this:
```ts
useForm({
  resolver: zodResolver(schema, {
    resolver: { mode: 'sync' },
    schema: schemaZodOptions
  })
})
```
I think I prefer this one, what do you think ?

FYI: `async` is the default mode

bluebill1049: This looks great: 

```
useForm({
  resolver: zodResolver(schema, options, {mode: 'sync'})
})
```

bluebill1049 (approved):

zod/src/types.ts
jorisre: `options` is optional so it could be:

```ts
useForm({
  resolver: zodResolver(schema, undefined, { mode: 'sync' })
})
```

PR #119, opened 2021-01-31
fix: expose array element validation errors in zod resolver

aprilrd:
### Issue
I noticed that the errors for array elements are not exposed in the Zod resolver. In the current version, the error looks like this instead (notice it doesn't expose the index of the validation error):
```
    "likedUsers": Object {
      "message": "Invalid input",
      "type": "invalid_union",
      "types": Object {
        "invalid_union": "Invalid input",
      },
    },
```

### Changes
- I updated the resolver to look at `unionErrors` when parsing errors from Zod to fix this error. Only `ZodInvalidUnionError` can have other errors as field values. So checking the existence of `unionErrors` should be good enough. But I am open to ideas here.

- I changed the original code's `reduce` to a queue to handle deeply nested union errors.

### DIFF
--- a/src/__snapshots__/zod.test.ts.snap
+++ b/src/__snapshots__/zod.test.ts.snap
@@ -133,6 +133,23 @@ Object {
         "invalid_type": "Expected number, received string",
       },
     },
+    "likedUsers": Object {
+      "0": Object {
+        "id": Object {
+          "message": "Expected number, received string",
+          "type": "invalid_type",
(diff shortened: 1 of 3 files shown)


jorisre (approved):
Good catch ! Thanks @aprilrd !

bluebill1049:
Thanks, @aprilrd!

aprilrd:
I appreciate the quick review and the release 😄

PR #128, opened 2021-02-03
Reduce superstruct resolver size

jorisre:
I removed `criteriaMode` because Superstruct can't return all errors for a same field. By removing `criteriaMode` the bundle's size is reduced by ~74%. We just have to add an explanation about Superstruct resolver in the new documentation for users.
Let me know if you're ok with these changes

I found an explanation about errors in typings:
carbon


superstruct/package.json
```diff
@@ -12,6 +12,7 @@
   "license": "MIT",
   "peerDependencies": {
     "react-hook-form": ">=6.6.0",
-    "@hookform/resolvers": "^1.0.0"
+    "@hookform/resolvers": "^1.0.0",
+    "superstruct": ">+0.12.0"
```
jorisre, line 16: For inlining superstruct

bluebill1049:
Oh, I didn't know about this. Thanks for digging into that. I will include that in the doc.

bluebill1049 (approved):

bluebill1049 (approved):

bluebill1049 (approved):

PR #130, opened 2021-02-05
Reduce yup resolver's size

jorisre:
I removed one test because it couldn't occur with an object schema. Let me know if I'm wrong


yup/package.json
```diff
@@ -11,6 +11,7 @@
   "types": "dist/index.d.ts",
   "license": "MIT",
   "peerDependencies": {
-    "react-hook-form": ">=6.6.0"
+    "react-hook-form": ">=6.6.0",
```
bluebill1049, line 14: FYI:I think with the latest update (i will move to 7) there are breaking changes.

bluebill1049:
> I removed one test because it couldn't occur with an object schema. Let me know if I'm wrong

👍

bluebill1049 (approved):
Awesome cleanup and bundle size reduction.

PR #131, opened 2021-02-05
Reduce joi resolver's size

jorisre:
- Reduce Joi resolver's size
- Reduce Vest resolver's size


bluebill1049:
wow, -25% awesome, and the code is so much cleaner as well!

bluebill1049 (approved):

PR #152, opened 2021-04-10
Feature/io-ts resolver

euphocat:
Hello everyone ! I've been using react-hook-form with fp-ts and io-ts since last summer on several projects, and I think I can share this resolver with confidence. This might not be the easiest tool (like the whole fp-ts world) but extremely powerful. Let me know what you think about it :)


io-ts/__tests__/__snapshots__/io-ts.ts.snap
```diff
@@ -0,0 +1,59 @@
...
+        "ref": undefined,
+        "type": "number",
+      },
+    ],
+  },
+  "values": Object {},
+}
+`;
+
+exports[`ioTsResolver should return all the errors from ioTsResolver when validation fails with \`validateAllFieldCriteria\` set to true 1`] = `
```
jorisre, line 20: When `validateAllFieldCriteria` is set to `true`, the error object contains a property `types` with all the errors, here is an example: https://github.com/react-hook-form/resolvers/blob/master/zod/src/__tests__/__snapshots__/zod.ts.snap#L258

io-ts/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "main": "dist/io-ts.js",
+  "module": "dist/io-ts.module.js",
+  "umd:main": "dist/io-ts.umd.js",
+  "source": "src/index.ts",
+  "types": "dist/index.d.ts",
+  "license": "MIT",
+  "peerDependencies": {
+    "react-hook-form": "^7.0.0",
+    "@hookform/resolvers": "^2.0.0",
+    "io-ts": "^2.2.13",
```
jorisre, line 16:
```suggestion
    "io-ts": "^2.0.0",
```

I'm new to `io-ts`, I read the changelog and I think it safes to use `^2.0.0` can you confirm? Same question for `fp-ts`

jorisre (commented):
Wow ! Another resolver ! 🎉 
Thanks a lot @euphocat 👏🏻 

> This might not be the easiest tool (like the whole fp-ts world) but extremely powerful.
You right 😅

README.md
```diff
@@ -298,6 +298,42 @@ const App = () => {
 export default App;
 ```
 
+### [io-ts](https://github.com/gcanti/io-ts)
+
+Validate your data with powerful decoders.
+
+[![npm](https://img.shields.io/bundlephobia/minzip/io-ts?style=for-the-badge)](https://bundlephobia.com/result?p=io-ts@2.2.16)
```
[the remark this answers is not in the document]
euphocat: 👍 I'll update this

io-ts/__tests__/__snapshots__/io-ts.ts.snap
euphocat: oh how interesting ! I've used the simple single, way but it's a good point

io-ts/package.json
euphocat: About `io-ts`, I don't use a lot of stuff from it, so yes, 2.0.0 should be way enough. Concerning `fp-ts`, I use `ReadonlyArray.last` for instance that was introduced in 2.5.0. I'll check it more carefully. Any thanks a lot for your feedbacks @jorisre !

bluebill1049:
another resolver! how exciting! Thanks for the PR and @jorisre's review.

README.md
```diff
@@ -298,6 +298,42 @@ const App = () => {
...
+
+```typescript jsx
+import React from 'react';
+import { useForm } from 'react-hook-form';
+import { ioTsResolver } from '@hookform/resolvers/io-ts';
+import t from 'io-ts';
+// you don't have to use io-ts-types but it's very useful
+import tt from 'io-ts-types';
```
bluebill1049, line 313: minor thoughts: perhaps we want to keep the example simple without introducing another lib?

euphocat: Yeah, I wondered about wether I should use it or not... `io-ts` by itself is very low-level. So I though "hum maybe it'll be scary for newcomers". But I can make it simpler for sure. It's just if you wan't a nice validator like `min` or `max` you have to craft it yourself !

bluebill1049: I don’t mind either 👌just a minor thought.

README.md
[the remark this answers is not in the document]
euphocat: done

io-ts/__tests__/__snapshots__/io-ts.ts.snap
euphocat: ok, it should be working as expected now

io-ts/package.json
euphocat: So it's ok for `io-ts` @ `^2.0.0` but for `fp-ts` I can only go to [^2.7.0](https://github.com/gcanti/fp-ts/releases/tag/2.7.0) as I use `ReadonlyArray.Foldable`

jorisre (approved):
Lgtm ! Thanks @euphocat

bluebill1049 (approved):
Thanks everyone!

euphocat:
thank you @jorisre and @bluebill1049 :)

Matthijn:
I'm not quite sure if this is the right section. But one thing is not quite clear to me.

Having this.

```
const schema = t.type({
  name: tt.NonEmptyString,
  description: tt.NonEmptyString
})

type FormValues = t.TypeOf<typeof schema>

export default function EditJob(job: JobDto) {
  const { register, handleSubmit, watch, formState: { errors } } = useForm<FormValues>({
    resolver: ioTsResolver(schema),
  });

  return (
      <main>
        <h1>Edit {job.name}</h1>

        <form onSubmit={handleSubmit(onSubmit)}>
          <div>
            <label htmlFor="name">Name:</label>
            <input type="text" {...register("name")} defaultValue={job.name} />
            <p>{errors.name?.message}</p>
          </div>
          <div>
            <label htmlFor="description">Description:</label>
            <textarea {...register("description")} defaultValue={job.description} />
            <p>{errors.description?.message}</p>
          </div>
          <input type="submit" value="Save" />
        </form>
      </main>
  )
}

```

On the lines where error messages can be displayed like `<p>{errors.name?.message}</p>` the property message doesn't exist with this resolver. (They do exist when I use other resolves like `yup`). 

The property type of `errors.name` seems to now be of type `DeepMap<tt.NonEmptyString, FieldError>` and I'm not quite sure how I can extract the error message to display from there.

jorisre:
@Matthijn Thank you for reporting the issue. Can you [open an issue](https://github.com/react-hook-form/resolvers/issues/new/choose) please ? And provide a codesandbox ? thank you

Matthijn:
It may be my mistake. Sorry. 

Doing this.

```
// type FormValues = t.TypeOf<typeof schema>
interface FormValues {
  name: string,
  description: string
}
```

Seems to work. Instead of the commented out part. (I thought that would prevent me from writing basically the same structure twice).

Issue #155, opened 2021-04-13
Yup resolver - missing errors when `criteriaMode: all`


jorisre:
https://codesandbox.io/s/react-hook-form-basic-forked-60bno?file=/src/index.js
React Hook Form - Basic (forked) - CodeSandbox
This is example with my problem.

Kind Regards
Jason

On 13 Apr 2021, 15:29 +0100, Joris ***@***.***>, wrote:
> Hi @codemaster115
> Can you share a codesanbox with your problem please ?
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.

_Originally posted by @codemaster115 in https://github.com/react-hook-form/resolvers/issues/26#issuecomment-818798500_

PR #156, opened 2021-04-14
fix(yupResolver): multiple errors messages with same type

jorisre:
Related to #155 

I'll update the v1 as well

### DIFF
--- a/yup/src/yup.ts
+++ b/yup/src/yup.ts
@@ -17,12 +17,17 @@ const parseErrorSchema = (
     }
 
     if (validateAllFieldCriteria) {
+      const types = previous[error.path!].types;
+      const messages = types && types[error.type!];
+
       previous[error.path!] = appendErrors(
         error.path!,
         validateAllFieldCriteria,
         previous,
         error.type!,
-        error.message,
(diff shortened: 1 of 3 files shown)


yup/src/yup.ts
```diff
@@ -17,12 +17,17 @@ const parseErrorSchema = (
...
+      const messages = types && types[error.type!];
+
       previous[error.path!] = appendErrors(
         error.path!,
         validateAllFieldCriteria,
         previous,
         error.type!,
-        error.message,
+        // eslint-disable-next-line
+        // @ts-expect-error
```
jorisre, line 29: Screenshot 2021-04-14 at 17 09 57

bluebill1049: I think you can do a cast here.

bluebill1049 (approved):
Awesome! thanks @jorisre

yup/src/yup.ts
jorisre: Updated, you can review again

bluebill1049 (approved):
❤️ awesome.

PR #157, opened 2021-04-15
Fix errors with criteria mode all

jorisre:
Fix other resolvers when `criteriaMode: all`


bluebill1049 (approved):
Thanks for fixing all! ❤️

PR #174, opened 2021-06-09
ci(dependabot): setup dependabot

jorisre:


dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
```
bluebill1049, line 6: `weekly`?

bluebill1049 (approved):
LGTM!

dependabot.yml
jorisre: Just to start and I'll change it later

bluebill1049: kk 👍

PR #181, opened 2021-07-06
Use zod schema inference to add types to callback

esemeniuc:


jorisre:
Thank you for your contribution @esemeniuc but the examples in the Readme are in plain JS (no TS)

esemeniuc:
I see, can we maybe add a comment for typescript users?

jorisre:
We already have a similar example here: https://react-hook-form.com/api/useform/#validationResolver
Scroll down a bit and select Zod instance and then TS.

esemeniuc:
Ah I did not see that page. Perhaps could you link the example from https://react-hook-form.com/get-started/#schemavalidation?

jorisre:
Sure! Do you want to add it at the top of the readme?

polarathene:
Slightly off-topic, but as someone new to TypeScript, Zod and RHF, what additional benefit does the type hint in the PR provide?

I've done a basic form prior with TS where the onSubmit method needed typing for the event otherwise it'd complain, but as that doesn't happen with the current Zod resolver example lacking the typing this PR adds, I'm curious what it helps with (`register`, `handleSubmit`, `formState.errors`?).

**EDIT:** I noticed it gave type hints to the string `register()` values and errors, which is useful.

---

I know that like with my basic form I still need to provide the type if the method isn't inlined, but the `useForm<type>` in this PR doesn't make a difference there:

```jsx
const handleData = (d: z.infer<typeof schema>) => console.log(d)
// ...
<form onSubmit={handleSubmit(handleData)}>
```

---

Regarding the two links, I did come across the `get-started` one, but wasn't aware of the separate resolvers package until I saw a `dev.to` blogpost about it when looking into validation. I get that it's imported as a separate package in the example, no issue there.

I would have appreciated the other mentioned validators having their examples from the main docs on that `get-started` page though. I followed the docs link mentioned in this thread, but found it very noisy and a bit difficult to know what I was looking at. 

The `resolvers` heading with typing to the right blended in the surrounding text above and below, both on light/dark themes, it didn't stand out as a heading and the examples required scrolling down further. On my browser, opening the link to a new tab had `shouldUseNativeValidation` heading near the top, few other lines preceding it.

Looking for something relevant I first noticed the "Rules" which was larger and in bold, and further down the actual examples being discussed. Most users would perhaps be visiting that docs link for the example snippets (I looked for this repo and saw examples on the README prior to knowing of the official docs examples), `get-started` would probably provide better UX. The heading is clear, and the examples snippets easily nearby to jump straight to.

jorisre:
@polarathene to get your `handleData` type-safe you can do the following:

```tsx
const handleData = handleSubmit((d) => console.log(d));
 
return (<form onSubmit={handleData}>)
```

jorisre:
Closing in favor of https://github.com/react-hook-form/resolvers/pull/184

PR #276, opened 2021-11-05
fix #275

### LINKED ISSUE #275: DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at node_modules\@hookform\resolvers\package.json
I've upgraded NextJS version to v12 and changed importing path of yupResolver.

```js
import { yupResolver } from "@hookform/resolvers/yup/dist/yup";
```

That seems like work well in dev, but I get this when I build to production.

image

JamesWong1999:
fix #275 with deprecate warning

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
       "require": "./typanion/dist/typanion.js"
     },
     "./package.json": "./package.json",
-    "./": "./"
+    "./*": "./*"
   },
   "files": [
     "dist",


bluebill1049 (approved):

bluebill1049:
thanks @JamesWong1999 for the PR, @jorisre any idea why this one fails?

jorisre:
It fails because `./*` is not supported by `check-exports-map`

bluebill1049:
thanks for the PR anyhow 👍

jorisre:
There is a similar issue on the Preact repo: https://github.com/preactjs/preact/issues/3299

brycefranzen:
Looks like check-exports-map has had an update.  Does this work now?  Could we merge that "closed" MR now to resolve this warning?

bluebill1049:
thanks for the contribution.

PR #284, opened 2021-11-15
patch(vestResolver): prepare for next major

ealush:
Vest's next major version that will come out soon ditches default exports to allow better tree shaking support and a simpler API. Named exports are already supported in the current version, in this PR I am just updating the docs and the internals to use named exports so they are future ready.


bluebill1049 (approved):
Awesome! looking forward to the next major version.

PR #327, opened 2022-01-18
Fix/yup silently fail

jorisre:


bluebill1049 (approved):
LGTM! perhaps this could be a minor update?

jorisre:
> LGTM! perhaps this could be a minor update?

yep, I'll open a PR for the v1 as well

PR #401, opened 2022-05-12
Add ability to skip zod's parsing

KATT:
When using [tRPC](https://trpc.io/), it's common to use the same validation schema on the server as the client. Therefore, it's good to be able to send the unparsed raw values from the client to the server.

Maybe this should be the default behavior, unsure if it's intentional to be as it is.

### DIFF
--- a/zod/src/__tests__/__snapshots__/zod.ts.snap
+++ b/zod/src/__tests__/__snapshots__/zod.ts.snap
@@ -13,6 +13,11 @@ Object {
       "ref": undefined,
       "type": "invalid_type",
     },
+    "dateStr": Object {
+      "message": "Required",
+      "ref": undefined,
+      "type": "invalid_type",
+    },
     "email": Object {
       "message": "Invalid email",
       "ref": Object {
@@ -83,6 +88,11 @@ Object {
     },
     "birthYear": Object {
       "message": "Expected number, received string",
+      "ref": undefined,
+      "type": "invalid_type",
+    },
+    "dateStr": Object {
+      "message": "Required",
       "ref": undefined,
       "type": "invalid_type",
     },
@@ -167,6 +177,14 @@ Object {
       "type": "invalid_type",
       "types": Object {
         "invalid_type": "Expected number, received string",
+      },
+    },
+    "dateStr": Object {
+      "message": "Required",
+      "ref": undefined,
+      "type": "invalid_type",
+      "types": Object {
+        "invalid_type": "Required",
       },
     },
     "email": Object {
@@ -284,6 +302,14 @@ Object {
         "invalid_type": "Expected number, received string",
       },
     },
+    "dateStr": Object {
+      "message": "Required",
+      "ref": undefined,
+      "type": "invalid_type",
+      "types": Object {
+        "invalid_type": "Required",
+      },
+    },
     "email": Object {
       "message": "Invalid email",
       "ref": Object {
@@ -373,5 +399,32 @@ Object {
     },
   },
   "values": Object {},
+}
+`;
+
+exports[`zodResolver should return parsed values from zodResolver with \`mode: sync\` when validation pass 1`] = `
+Object {
+  "errors": Object {},
+  "values": Object {
+    "accessToken": "accessToken",
+    "birthYear": 2000,
+    "dateStr": 2020-01-01T00:00:00.000Z,
+    "email": "john@doe.com",
+    "enabled": true,
+    "like": Array [
+      Object {
+        "id": 1,
+        "name": "name",
+      },
+    ],
+    "password": "Password123_",
+    "repeatPassword": "Password123_",
+    "tags": Array [
+      "tag1",
+      "tag2",
+    ],
+    "url": "https://react-hook-form.com/",
+    "username": "Doe",
+  },
 }
 `;
--- a/zod/src/zod.ts
+++ b/zod/src/zod.ts
@@ -69,7 +69,7 @@ export const zodResolver: Resolver =
 
         return {
           errors: {} as FieldErrors,
-          values: data,
+          values: resolverOptions.rawValues ? values : data,
         };
       } catch (error: any) {
         return {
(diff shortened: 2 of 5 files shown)


zod/src/zod.ts
```diff
@@ -69,7 +69,7 @@ export const zodResolver: Resolver =
 
         return {
           errors: {} as FieldErrors,
-          values: data,
+          values: resolverOptions.rawValues ? values : data,
```
KATT, line 72: Unsure if it was ever intentional to pass the parsed output here.

bluebill1049: Thanks for the PR, it's intentional to return what's provided by the resolvers. What's the issue that brings?

KATT: I made a reproduction


1. Go to https://kitchen-sink-git-repro-trpc.vercel.app/react-hook-form
2. Open network tab
3. Fill in form

- Expected: submits successfully
- Actual doesn't work


Highlighted inline here https://github.com/trpc/examples-kitchen-sink/pull/36 on why it doesn't work - it's because the zod resolver coerces the values before sending them to the server which will then try to validate them based on the input:

1. `dateStr` gets validated as `Date` on client
2. `Date` is sent to server
3. Server receives `Date`, tries to validate as a string --> ❌

bluebill1049: thanks for the context. `rawValues` option I couldn't find in zod's doc do you have a reference?

KATT: With raw just means that we're not using zod's parsed output

bluebill1049: yea, is that a public API? couldn't find it under zod readme.

KATT: It has nothing to do with zod, it's just an option to opt-out of the zod transforms within the form. I don't think they should've been there in the first place as zod's `.transform()` can change the data type

Did this as a further explanation: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwF4F84DMoRHA5EiAE1wFgAoc8gYwgDsBneeqgCwFMQBDOAXjiQB0EAEYArNlRgAKBOTjy4hTjDYBlGFABc-OQoGMowWgHMpASl3yBGzg3TQQUqQDdOAGwCubM7wB8cWjYAdzgAEWU2F3cvMwsyBTgBKDZ0I0ioz28-OABCYHoAOU4CjK8BYzYYABVQSNiAGkRLBRA2enpOCu1cAElaVzdgQkUI3HrmlDNxsknKMhoGeCMwD3g+WXiRlXUtOAAiACYABmOAWiOARnOLvemUAG45hcY4ZPpeOGZ2Ln1OdDYAAqcKD0SLLVZxObAdBwKRvfQeKhUNr0HwbeQAegxcCqLHycHxMHY-CIrzaEDczjYUAJ7w89CMxlewGMLHgtAgQV0zwpbAEbggpnhShgnDimOxCSlUoAegB+XRYuA9eBBYBuNxwGx2Bxa4kGRkE2gwCBwbjhFTcuj0Xn8wVwtoCEWcJ0RHbiuBK6Xe6XyuYoOZKtVEuAQDw0qCcIIANWibVDYBgwGtcHYyUUEHjhNNRkYbE4w3pbDgAANwTAS9RrbaBaZy3EvT64H6yEqABLU4tE4sitQaMlcXNwYRsQ3cA0mOBBPGsAlLd4j+zp7tayM6qAgKsMGv28uu7YaBuSpuyuVAA

zod/src/__tests__/__snapshots__/zod.ts.snap
```diff
@@ -375,3 +401,30 @@ Object {
...
 }
 `;
+
+exports[`zodResolver should return parsed values from zodResolver with \`mode: sync\` when validation pass 1`] = `
+Object {
+  "errors": Object {},
+  "values": Object {
+    "accessToken": "accessToken",
+    "birthYear": 2000,
+    "dateStr": 2020-01-01T00:00:00.000Z,
```
KATT, line 411: You see here @bluebill1049, the default behavior is to apply the transforms which will give a `Date` here and not a string

zod/src/zod.ts
bluebill1049: sorry, I am blind... I saw you added a new config.

bluebill1049: If we are adding this new config, it probably should apply to all resolvers, which I am happy to look into this and get documented as well.

bluebill1049: Thanks, @KATT I think the config name is great. 👍

KATT: I had a look around and there's inconsistency how the resolvers behave in terms of this

Here's a few that return the raw value as default:

- https://github.com/react-hook-form/resolvers/blob/350d255817de8820aa62e6f2036daff211708d1e/vest/src/vest.ts#L50
- https://github.com/react-hook-form/resolvers/blob/350d255817de8820aa62e6f2036daff211708d1e/typanion/src/typanion.ts#L37
- https://github.com/react-hook-form/resolvers/blob/350d255817de8820aa62e6f2036daff211708d1e/nope/src/nope.ts#L45
- https://github.com/react-hook-form/resolvers/blob/350d255817de8820aa62e6f2036daff211708d1e/class-validator/src/class-validator.ts#L61

KATT: If you want to avoid breaking changes, we can't have a default behavior for this. I think it's better to only add it to zod for now.

bluebill1049: great find 👍 let's keep them consistent. ❤️

KATT: 1. Do you want this to be done ahead of merging this PR?
2. What behavior do you want as default?

bluebill1049: The default behavior should return by what returned by the resolver 

```tsx
rawValues: false
```

and users can opt in to return rawValues if they want the values to return by hook form. I am happy to merge this PR to get started just let @jorisre do a review on this as well.

bluebill1049 (approved):
LGTM

jorisre (approved):
LGTM ! Thank you for the PR @KATT

bluebill1049:
@jorisre merge when you feel comfortable ❤️

PR #412, opened 2022-06-10
fix: ajv resolver to work with unlimited layers of nesting

NoRKin:
The current implementation of `ajv` resolver didn't support well more than 2 layers of nesting. The error keys wouldn't match the shape of the data. See below `data/category`.

The below schema results in:

```js
{
  type: 'object', 
  properties: {
    row: {
      type: 'object',
      properties: {
	name: {
	  type: 'string',
	  minLength: 10,
	  errorMessage: { minLength: 'Username should be at least 10 characters' },
	},
	data: {
	  type: 'object',
	  properties: {
	    category: {
	      type: 'string',
	      minLength: 1,
	      errorMessage: { minLength: 'Category is required' },
	    }
	  },
	  required: ['category'],
	  additionalProperties: true,
	}
      },
      required: ['name', 'data'],
      additionalProperties: true,
    },
  },
  required: ['row'],
  additionalProperties: false,
};

```

Screen Shot 2022-06-10 at 1 18 20 pm
 
The fix is to essentially replace all occurrences of `/` instead of only the first one.

### DIFF
--- a/ajv/src/__tests__/__fixtures__/data.ts
+++ b/ajv/src/__tests__/__fixtures__/data.ts
@@ -4,7 +4,7 @@ import { Field, InternalFieldName } from 'react-hook-form';
 interface Data {
   username: string;
   password: string;
-  deepObject: { data: string };
+  deepObject: { data: string; twoLayersDeep: { name: string } };
 }
 
 export const schema: JSONSchemaType<Data> = {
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
thanks for the PR 🙏

jorisre (approved):

PR #419, opened 2022-06-23
Prefer `fp-ts/Either` over `fp-ts/lib/Either`

ryota-ka:
Unlike other import statements, this one points not to `fp-ts/Either` but to `fp-ts/lib/Either`, which is a CommonJS file (`node_modules/fp-ts/lib/Either.js`).
This causes an error when you import `@hookform/resolvers/io-ts` from a ES module, where a file extension is mandatory in import sources, as `fp-ts/lib/Either` lacks `.js` extension.
Importing from `fp-ts/Either` (just as all other imports do so) solves the problem, since the `package.json` switches CJS/ESM.


bluebill1049:
thanks for contributing.

Issue #423, opened 2022-06-28
Private packages such as "ajv" and "class-validator" shouldn't share the name of official npm packages. This causes false positives with security scans, among other issues.


Fallibilist:
These private packages should probably be renamed to something like "ajv-wrapper" or "ajv-plugin" to avoid being confused for the official package.

As an example "ajv" versions before v6.12.3 have a vulnerability (detailed here: https://nvd.nist.gov/vuln/detail/CVE-2020-15366).

The private "ajv" package included in the project simply has a version of "1.0.0"
![image](https://user-images.githubusercontent.com/29155351/176254752-4d7b1f67-ef03-4ff5-89b0-94afac49f01d.png)

Security scans find a false positive here because the name of the package is identical to the official "ajv" package, even though the actual package is >= v6.12.3
![image](https://user-images.githubusercontent.com/29155351/176255903-1a63c912-6943-4eab-a1a8-a71041243feb.png)
![image](https://user-images.githubusercontent.com/29155351/176255377-6cede584-06a6-459f-b05c-58058554353c.png)

Can these private packages be renamed to someone more descripting which doesn't match an official package?

Thanks!

bluebill1049:
maybe rename them to `@hookform/resolvers/ajv` ? feel free to send a pr on this.

Fallibilist:
Thanks @bluebill1049 !
I made a PR here: https://github.com/react-hook-form/resolvers/pull/424

jorisre:
Fix in #424

PR #424, opened 2022-06-30
Renamed private packages to avoid ambiguity with existing official packages. This should mitigate false positives during image scans.

Fallibilist:
This PR addresses the issue raised here: https://github.com/react-hook-form/resolvers/issues/423

This change was pretty minor but I did have a couple of questions:
- Would you prefer if I changed the path for these dependencies as well? I decided against it because it might be a breaking change but I can update the PR with that if you'd like.
- I get a warning about the new package names due to the double forward slashes (/) failing the regex. It doesn't have an impact since these are private packages but I wasn't sure if you would rather I use a different naming convention. Something like "@hookform/resolvers.ajv" perhaps.
image

### DIFF
--- a/ajv/package.json
+++ b/ajv/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "ajv",
+  "name": "@hookform/resolvers/ajv",
   "amdName": "hookformResolversAjv",
   "version": "1.0.0",
   "private": true,
--- a/class-validator/package.json
+++ b/class-validator/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "class-validator",
+  "name": "@hookform/resolvers/class-validator",
   "amdName": "hookformResolversClassValidator",
   "version": "1.0.0",
   "private": true,
--- a/computed-types/package.json
+++ b/computed-types/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "computed-types",
+  "name": "@hookform/resolvers/computed-types",
   "amdName": "hookformResolversComputedTypes",
   "version": "1.0.0",
   "private": true,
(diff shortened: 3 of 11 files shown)


jorisre:
I can't test right now but I'm sure deep import like @hookform/resolvers/zod still work if we rename the package name

bluebill1049:
> I can't test right now but I'm not sure deep import like @hookform/resolvers/zod still work if we rename the package name

I see, that's a valid concern. @Fallibilist any chance you can valid this above?

Fallibilist:
I ran a few tests and this seems to be fine. It looks like the path in exports is all that matters (in this case ./zod).

Here is how I tested.
I created a simple testing project which imported @hookform/resolvers from a local directory. It imported zod and ran simple schema validation.
image
image
image

This is successful. 
image

Now if I rename the export "./zod" to "./zodbroken" it breaks
image
image

If I set the import to this value then it works again.
image


This was all conducted with the zod resolver's package name as "@hookform/resolvers/zod"

Here are those two files if you want to test yourself:
`{
  "name": "export-test",
  "version": "1.0.0",
  "description": "",
  "main": "testExports.js",
  "type": "module",
  "scripts": {
    "start": "node testExports.js",
    "zod": "^3.11.6"
  },
  "dependencies": {
    "@hookform/resolvers": "./resolvers"
  }
}`

`import { zodResolver } from '@hookform/resolvers/zod';
import * as z from 'zod';

const schema = z.object({
  name: z.string().min(1, { message: 'Required' }),
  age: z.number().min(10),
});

console.log(await zodResolver(schema)({name: 'Name', age: 5}, {}, {}));
`

jorisre:
Thanks a ton for your contribution ! Looks good to me :)

PR #425, opened 2022-07-06
Fix nested schema with computed-types

hlubek:
Having nested fields (represented by a nested `Schema` in computed-types) did not produce the expected error messages.

This change fixes it by using paths with dot syntax as the errors key, which will be correctly converted by `toNestError`.


bluebill1049 (approved):
LGTM

jorisre (approved):
Thank you !

PR #427, opened 2022-07-13
fix(yup): fix Yup wrong import statement assuming default export

Grsmto:
Hi,

Yup was imported with a default import [but in the original library no default export is defined](https://github.com/jquense/yup/blob/master/src/index.ts). This is only working at the moment when building the lib because of the tsconfig option `esModuleInterop`.

I believe this is causing an issue only in specific scenarios (I had the issue while using a Webpack 4 setup with an older ts-loader version) but since it's a tiny change, might be reasonable to fix.

This is the error you could eventually see and this PR resolve:
`TS1192: Module '"/node_modules/@hookform/resolvers/yup/src/yup"' has no default export.`

### DIFF
--- a/yup/src/yup.ts
+++ b/yup/src/yup.ts
@@ -1,4 +1,4 @@
-import Yup from 'yup';
+import * as Yup from 'yup';
 import { toNestError, validateFieldsNatively } from '@hookform/resolvers';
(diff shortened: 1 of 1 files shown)


jorisre (approved):
Good catch ! thank you :)

PR #430, opened 2022-07-18
[AJV] Add option to run ajv with ajv-formats

jorgebarredo:
Add option to extend the ajv resolver with [ajv-formats](https://ajv.js.org/packages/ajv-formats.html)

[ajv-formats](https://github.com/ajv-validator/ajv-formats) makes it easy and secure to validate *emails*, *urls*, *uuids*, and many others string formats.

### DIFF
--- a/ajv/src/__tests__/__fixtures__/data.ts
+++ b/ajv/src/__tests__/__fixtures__/data.ts
@@ -7,6 +7,10 @@ interface Data {
   deepObject: { data: string; twoLayersDeep: { name: string } };
 }
 
+interface Email {
+  email: string;
+}
+
 export const schema: JSONSchemaType<Data> = {
   type: 'object',
   properties: {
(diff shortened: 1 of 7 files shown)


jorisre:
Hi 👋🏻 thank you for your PR. We can't add a such option because it's inconsistent across all of the resolvers. `factoryOtions` is not designed for this need.

@bluebill1049  what do you think?

jorgebarredo:
Hi. Thanks for the response.
I looked at other resolvers and some of them have `rawValues?: boolean`, so it is not that consistent. It is an optional property anyway.

In any case, I'm happy to consider other approaches. Do you have a recommendation for my case?

Thanks.

bluebill1049:
yea, unless it's something generic across all the schemas. The point of this lib is to create a standard for all custom resolvers, if this causes an issue for you, maybe it's worth creating a custom resolver.  Yes, `rawValues` is optional and we want to apply to all resovlers so the result will returned before the transformation.

jorisre:
It's something possible but I prefer to open an issue and wait to see if others users have the same need

jorgebarredo:
Thanks guys,
- I'm closing this PR in favour of: https://github.com/react-hook-form/resolvers/pull/431
- I also created an issue as agreed: https://github.com/react-hook-form/resolvers/issues/432

PR #442, opened 2022-08-11
Fix nested error reporting

### LINKED ISSUE #441: Yup errors are swallowed
**Describe the bug**
When a field has multiple errors, just the one higher in the hierarchy is reported, despite setting `criteriaMode: 'all'` and `earlyAbort: false`.

**To Reproduce**

1. Open the [sandbox](https://codesandbox.io/s/modest-butterfly-7noyn0?file=/src/App.tsx) 
2. Click on Submit
3. Open the console
4. Notice there is a single error for the items field: "items field must have at least 2 items"
5. Now, uncomment line 45
6. Reload the page
7. See it now reports errors on items that should have been reported along with the minimum number of items requirement

https://codesandbox.io/s/modest-butterfly-7noyn0?file=/src/App.tsx

**Expected behavior**
All errors.

marcospassos:
Fixes #441

### DIFF
--- a/src/__tests__/__snapshots__/toNestObject.ts.snap
+++ b/src/__tests__/__snapshots__/toNestObject.ts.snap
@@ -10,25 +10,22 @@ Object {
     },
     "type": "st",
   },
-  "test": Object {
-    "0": Object {
+  "test": Array [
+    Object {
       "name": Object {
         "message": "second message",
         "ref": undefined,
         "type": "nd",
       },
     },
-    "1": Object {
+    Object {
       "name": Object {
         "message": "third message",
         "ref": undefined,
         "type": "foo",
       },
     },
-    "message": "second message",
-    "ref": undefined,
-    "type": "st",
-  },
+  ],
 }
 `;
 
@@ -70,6 +67,13 @@ Object {
         "message": "second message",
         "ref": undefined,
         "type": "nd",
+      },
+    },
+    Object {
+      "name": Object {
+        "message": "third message",
+        "ref": undefined,
+        "type": "foo",
       },
     },
   ],
--- a/src/__tests__/__snapshots__/toNestObject.ts.snap
+++ b/src/__tests__/__snapshots__/toNestObject.ts.snap
@@ -10,15 +10,25 @@ Object {
     },
     "type": "st",
   },
-  "test": Array [
-    Object {
+  "test": Object {
+    "0": Object {
       "name": Object {
         "message": "second message",
         "ref": undefined,
         "type": "nd",
       },
     },
-  ],
+    "1": Object {
+      "name": Object {
+        "message": "third message",
+        "ref": undefined,
+        "type": "foo",
+      },
+    },
+    "message": "second message",
+    "ref": undefined,
+    "type": "st",
+  },
 }
 `;
 
--- a/src/toNestError.ts
+++ b/src/toNestError.ts
@@ -19,7 +19,7 @@ export const toNestError = <TFieldValues>(
           return 0;
         }
 
-        return leftPath > rightPath ? 1 : -1
+        return leftPath > rightPath ? -1 : 1
       });
 
   const fieldErrors = {} as FieldErrors<TFieldValues>;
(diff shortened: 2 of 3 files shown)


src/__tests__/__snapshots__/toNestObject.ts.snap
```diff
@@ -10,15 +10,25 @@ Object {
     },
     "type": "st",
   },
-  "test": Array [
-    Object {
+  "test": Object {
+    "0": Object {
```
bluebill1049, line 14: Why has this becoming object?

bluebill1049 (commented):
Thanks for the pr.

src/__tests__/__snapshots__/toNestObject.ts.snap
marcospassos: I believe it's because the snapshot serializer serializes arrays with non-numeric properties as objects.

marcospassos: That's because the [test error is an object](https://github.com/react-hook-form/resolvers/pull/442/files#diff-e8bad6d3d67175a94e94db1c8258febe5a4017bfa4908c1948a58e72570b07cfR6).

marcospassos:
@bluebill1049, is there anything else you'd like to address on this PR?

bluebill1049:
I am quite busy at the moment, but I will find some time to review it properly in the following days. thanks for the help <3

src/toNestError.ts
```diff
@@ -13,14 +13,29 @@ export const toNestError = <TFieldValues>(
...
   options.shouldUseNativeValidation && validateFieldsNatively(errors, options);
 
+  const entries = Object.entries(errors)
+      .sort(([leftPath], [rightPath]) => {
+        if (leftPath === rightPath) {
+          return 0;
+        }
+
+        return leftPath > rightPath ? -1 : 1
+      });
```
bluebill1049, line 23: This logic worries me because this function is a generic function and this will impact other resolvers as well.

bluebill1049 (commented):
I think **parent** error will swallow children's errors for now and keep things simple, maybe you can keep this resolver at your app level or even publish it, so others can reuse it as well. Again, thanks very much for looking into this and even providing a pr for it. 🙏

marcospassos:
The problem is that there is no easy way to fix or monkey patch this other than maintaining a fork. Also, this is a real feature bug for those who use the `all` criteria.

georgekaran:
I was having the same error on my application and I can confirm this PR solves the issue. Is there something missing so we can have this in a future release of resolvers?

marcospassos:
@georgekaran , this is a terrible but functional workaround for the issue:

```ts
import {ValidationError} from "yup";
import {yupResolver as baseResolver} from "@hookform/resolvers/yup";
import BaseSchema from "yup/lib/schema";

/**
 * Workaround for https://github.com/react-hook-form/resolvers/issues/441
 */

function fixErrorNesting(errors: Record<string, any>): Record<string, any> | any[] {
  const fieldErrors: Record<string, any> = {};

  for (const [key, value] of Object.entries(errors)) {
    if (typeof value !== 'object' || value === null) {
      fieldErrors[key] = value;

      continue;
    }

    if (Array.isArray(value)) {
      fieldErrors[key] = value.map(fixErrorNesting);

      continue;
    }

    if (!('0' in value)) {
      fieldErrors[key] = fixErrorNesting(value);

      continue;
    }

    const copy = {...value}
    const list = [];

    let index = 0;

    while (index in copy) {
      list.push(fixErrorNesting(copy[index]));
      delete copy[index];
      index++;
    }

    Object.assign(list, copy);

    fieldErrors[key] = list;
  }

  return fieldErrors;
}

export const yupResolver: typeof baseResolver = (schema, schemaOptions = {}, resolverOptions = {}) => {
  const schemaClone = (schema as any).clone() as BaseSchema;

  const validate = schemaClone.validate.bind(schemaClone);
  const validateSync = schemaClone.validateSync.bind(schemaClone);

  function fixError<T>(error: T): T {
    if (error instanceof ValidationError) {
      error.inner.sort((left, right) => {
        if (left.path === right.path) {
          return 0;
        }

        return (left.path ?? '') > (right.path ?? '') ? 1 : -1
      });
    }

    return error;
  }

  schemaClone.validate = async (...args) => {
    try {
      return await validate(...args);
    } catch (error) {
      throw fixError(error);
    }
  };

  schemaClone.validateSync = (...args) => {
    try {
      return validateSync(...args);
    } catch (error) {
      throw fixError(error);
    }
  };

  const resolver = baseResolver(schemaClone as typeof schema, schemaOptions, resolverOptions);

  return async (values, context, options) => {
      const result = await resolver(values, context, options);

      return {
        ...result,
        errors: fixErrorNesting(result.errors) as typeof result.errors,
      }
  };
};
```

It should definitely be fixed.

georgekaran:
Thanks @marcospassos it worked like a charm, but it should definitely be shipped native 😕

PR #498, opened 2023-01-28
Fix ts-io resolver error reporting for exact(intersection(...))

Firfi:
`exact(intersection(...))` structure would result in an undesirable result of a shape`{0: errorsRecord, 1: errorsRecord, ...}` whereas just exact(...) or just intersection(...) would correctly report errors of shape `errorsRecord`. 

This PR adds a test against it and resolves the bug.

UPD: same for Branded types

Bonus: caniuse-lite bumped a bit in package-lock


bluebill1049 (approved):
Thanks for the fix. LGTM

PR #504, opened 2023-02-03
Throw any error unrelated to Zod

stephan281094:
This pull request makes sure that errors that are not of type `ZodError` are thrown. It's possible for errors other than `ZodError` to occur when using things like `.refine` (for instance by referencing the property of a non-existent item in an array). Currently, when this happens, you get the following error:
>TypeError: undefined is not an object (evaluating 'e.length')

This took me a long time to debug today, because it was not clear where this error was coming from, nor was it very descriptive. Hopefully this change will save someone's time in the future!

Please let me know if any changes are necessary. I'll gladly update the pull request.

### DIFF
--- a/zod/src/__tests__/zod.ts
+++ b/zod/src/__tests__/zod.ts
@@ -77,4 +77,16 @@ describe('zodResolver', () => {
 
     expect(result).toMatchSnapshot();
   });
+
+  it('should throw any error unrelated to Zod', async () => {
+    const schemaWithCustomError = schema.refine(() => {
+      throw Error('custom error');
+    });
+    const promise = zodResolver(schemaWithCustomError)(validData, undefined, {
+      fields,
+      shouldUseNativeValidation,
+    });
+
+    await expect(promise).rejects.toThrow('custom error');
+  });
 });
--- a/zod/src/zod.ts
+++ b/zod/src/zod.ts
@@ -3,7 +3,7 @@ import {
   FieldError,
   FieldErrors,
 } from 'react-hook-form';
-import { z } from 'zod';
+import { z, ZodError } from 'zod';
 import { toNestError, validateFieldsNatively } from '@hookform/resolvers';
 import type { Resolver } from './types';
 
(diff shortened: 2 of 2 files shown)


jorisre:
Error should be associated to a field (by its name). That's how work resolvers with react-hook-form. 
@bluebill1049 what's your thoughts?

bluebill1049:
Yes, @jorisre let's keep it that way unless we have a general implementation for all resolvers. This would create inconsistency between all.

stephan281094:
> Error should be associated to a field (by its name). That's how work resolvers with react-hook-form.
> @bluebill1049 what's your thoughts?

Thank you for the quick reply! I'm not sure I understand. Are you saying instead of checking whether an error is of type `ZodError`, it should check for a different type of error?

stephan281094:
> > Error should be associated to a field (by its name). That's how work resolvers with react-hook-form.
> > @bluebill1049 what's your thoughts?
> 
> Thank you for the quick reply! I'm not sure I understand. Are you saying instead of checking whether an error is of type `ZodError`, it should check for a different type of error?

Let me know what changes are required for this to get through, or whether you want to close this, thank you!

jorisre:
@stephan281094 resolvers return errors that are linked to a field by its path. In your case, your refine isn't linked to a field. 
The correct way to use refine with react-hook-form is to put manually a path:

https://zod.dev/?id=customize-error-path

```ts
schema.refine((obj) => obj.password === obj.repeatPassword, {
    message: 'Passwords do not match',
    path: ['confirm'], // 👈 reference the field
  })
```

stephan281094:
> @stephan281094 resolvers return errors that are linked to a field by its path. In your case, your refine isn't linked to a field. The correct way to use refine with react-hook-form is to put manually a path:
> 
> https://zod.dev/?id=customize-error-path
> 
> ```ts
> schema.refine((obj) => obj.password === obj.repeatPassword, {
>     message: 'Passwords do not match',
>     path: ['confirm'], // 👈 reference the field
>   })
> ```

Thanks for clarifying, Joris. I understand that errors thrown when `z.refine` returns `false` should be linked to fields. I'm however not talking about Zod errors, I'm talking about non-Zod errors. Right now, the code always expects a ZodError to be thrown, when different kinds of errors may also be thrown.

For example, the following piece of code:
```ts
schema.refine((obj) => {
  return obj.non.existing.property
})
```
throws:
>TypeError: undefined is not an object (evaluating 'e.length')

when it should be throwing:
>TypeError: undefined is not an object (evaluating 'obj.non.existing')

The Zod resolver currently, incorrectly assumes the error that gets caught, is **always** a Zod Error. In the code you can see `error.isEmpty` being accessed, which doesn't exist on non-Zod Error objects. This leads to `parseErrorSchema` being called with the non-existing `error.errors` object, resulting in an incorrect error message.

I hope I've been able to better clarify the issue this pull request is solving.

jorisre:
That's correct, we do something similar for the yup resolver. Can you fix conflicts? Thank you

jorisre:
Here is #528 that fix all resolvers

jorisre:
Cloising in favor of #528

stephan281094:
> Here is #528 that fix all resolvers

Amazing, thank you! 🙌

PR #523, opened 2023-03-17
perf(ajvResolver): reduce bundle size

jorisre:


bluebill1049 (approved):
ooh wow almost 100% save, awesome!

ilteoood:
@bluebill1049 this is a breaking change that should be mentioned in the changelog

jorisre:
@ilteoood Can you elaborate? There is no breaking changes here. It's mentioned in the release note

image

ilteoood:
IMO this should be listed in the breaking changes, as it requires some actions to be done on the user side in order to upgrade smoothly.

If it was just a "performance improvement", I should be able to upgrade without doing any action, right? But this seems not to be the case, since now I need to include both `ajv` and `ajv-errors` as dependencies in my project.

jorisre:
Alright, I'll update the release notes. Just wanted to clarify that `ajv` and `ajv-errors` were not intended to be bundled by resolvers; it was a misconception from the start. On the bright side, this means you're free to use the specific versions of `ajv` and `ajv-errors` that you prefer!

ilteoood:
@jorisre you are completely right, but as this has been changed it should be explicitly written in the changelog.

Just to warn the user that he should take care of this starting now 😄

jorisre:
Done ✅

PR #524, opened 2023-03-17
feat(classValidationResolver): add transform and validator options

### LINKED ISSUE #445: class-validator resolver not working properly when using Expose annotations with groups
**Describe the bug**
When a class is annotated with `@Expose({groups:  "group-name"})` the validation doesn't work.
With the plain `@Expose()` annotation there's no problem.

**To Reproduce**
Create a simple class to validate:
```typescript
class TestDto {
  @Expose( {groups: ["find", "create", "update"]} )
  @Type(() => Number)
  @IsDefined({
    message: `All fields must be defined.`,
    groups: ["publish"],
  })
  @IsNumber({}, { message: `Must be a number`, always: true })
  @Min(0, { message: `Cannot be lower than 0`, always: true })
  @Max(255, { message: `Cannot be greater than 255`, always: true })
  random: number;
}
```

Then just create a simple resolver:
```typescript
const saveResolver = classValidatorResolver(TestDto , {
  groups: ["update"],
  skipMissingProperties: true,
});
``` 
All the validations are ignored.

My assumption is that the resolver, transforms the value into an instance of the TestDto class without any particular option to expose the given annotated properties. So the resulting instance doesn't have the property. So as `skipMissingProperties: true` is  set, the validations won't run on it, except for the `@IsDefined` annotation if it was annotated with the same group as the resolver: `@IsDefined({groups: ["update"]})`.

**Expected behavior**
The validation should be run on the object, and reject it if it doesn't meet the criteria.

jorisre:
@bluebill1049 you can review but do not merge. There is a breaking change

Fix #445

### DIFF
--- a/class-validator/src/__tests__/__snapshots__/class-validator.ts.snap
+++ b/class-validator/src/__tests__/__snapshots__/class-validator.ts.snap
@@ -205,3 +205,38 @@ exports[`classValidatorResolver > should return all the errors from classValidat
   "values": {},
 }
 `;
+
+exports[`validate data with transformer option 1`] = `
+{
+  "errors": {
+    "random": {
+      "message": "All fields must be defined.",
+      "ref": undefined,
+      "type": "isDefined",
+      "types": {
+        "isDefined": "All fields must be defined.",
+        "isNumber": "Must be a number",
+        "max": "Cannot be greater than 255",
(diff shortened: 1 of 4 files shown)


bluebill1049 (approved):
👍

Issue #471, opened 2022-11-19
[ajv] Transform data before validation


elton-okawa:
**Is your feature request related to a problem? Please describe.**
I have a dynamically generated form based on a JSON Schema and I’m facing problems when we have `optional` fields with validations, as every registered input has its own key on the form values, they are validated and fails even though visually they are empty.

[CodeSandbox example](https://codesandbox.io/s/react-hook-form-optional-fields-w1bo3k?file=/src/index.js)
In this example, we have an optional field `description` which, if defined, needs to have a minimum length of 5 characters.
From the user perspective, an optional field can be left empty but it doesn’t work here because the `description` field exist on the submitted values and consequently it’s validated.
Another example might happen with phone number or email, they might be optional but if they are provided, they must follow a pattern.

**Describe the solution you'd like**

Allow the `ajvResolver` receive an function to transform the form values before validation.
I like this solution because:
- there is no magic happening behind the scenes (the user will define the function)
- we can use the same function before submitting the data

For example:
```jsx
function transform(values) {
  // clone and modify form values before validation, like removing empty fields
  return modifiedValues;
}

const { handleSubmit }= useForm({
  resolver: ajvResolver(schema, transform)
});

<form onSubmit={handleSubmit((data) => console.log(transform(data))}>
  ...
</form>
```

**Describe alternatives you've considered**

I have considered the following options:

1. Button to hide and show the optional field
    - I could leverage the unregister on unmount feature as unregistered fields can be removed from the form values
    - It works but I think the UX would become a bit worse as every optional fields would need to click a button to be filled
2. Custom controller to register and unregister field based on input value
    - register if `input !== ''` and unregister otherwise
    - It also works but with some concerns
        - unregistering a field made the entire form re-render (not sure if it’s the expected behavior or I’m doing something wrong)
        - I felt that I was going further on a fragile solution (not the expected way to use the library)

**Additional context**

Kind of related to [react-hook-form #9278](https://github.com/react-hook-form/react-hook-form/issues/9278) but for `ajv` validations
For now I customized the official resolver like the solution I’ve described, I thought that I might be useful to someone else and I could receive feedback about the way that I’m solving the problem.

I’m also willing to open up a PR if it seems a good solution 🤓

jorisre:
@elton-okawa Do you find a way to manage this?

elton-okawa:
@jorisre I basically implemented my suggested solution, I'm planning to open up a PR with some usage documentation this week.

PR #533, opened 2023-03-21
Feature/ajv transform before validation

elton-okawa:
Solves #471

### DIFF
--- a/ajv/src/ajv.ts
+++ b/ajv/src/ajv.ts
@@ -1,11 +1,11 @@
 import { toNestError, validateFieldsNatively } from '@hookform/resolvers';
-import Ajv, { DefinedError } from 'ajv';
+import Ajv, { ErrorObject, ValidateFunction } from 'ajv';
 import ajvErrors from 'ajv-errors';
 import { appendErrors, FieldError } from 'react-hook-form';
 import { Resolver } from './types';
 
 const parseErrorSchema = (
-  ajvErrors: DefinedError[],
+  ajvErrors: ErrorObject[],
   validateAllFieldCriteria: boolean,
 ) => {
   // Ajv will return empty instancePath when require error
@@ -46,7 +46,7 @@ const parseErrorSchema = (
 };
 
 export const ajvResolver: Resolver =
-  (schema, schemaOptions, { mode = 'sync', transform } = {}) =>
+  (schema, schemaOptions, { mode = 'async', transform } = {}) =>
   async (values, _, options) => {
     const ajv = new Ajv(
       Object.assign(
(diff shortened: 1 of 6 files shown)


ajv/src/ajv.ts
```diff
@@ -63,12 +63,12 @@ export const ajvResolver: Resolver =
 
     const validate = ajv.compile(
       Object.assign(
-        { $async: resolverOptions && resolverOptions.mode === 'async' },
+        { $async: mode === 'async' },
```
elton-okawa, line 66: Ajv resolver was not strictly following the root docs that states that default mode is `async`

I kept the default value as `sync` in order to not change actual behavior for now

jorisre (changes requested):
Thank for the PR, we really appreciate it. Well tested and documented 🚀 
I would like to avoid introduce a breaking change. Is there any other way to transform data with AJV?
 
Waiting for @bluebill1049 review.

ajv/src/ajv.ts
elton-okawa: I totally agree with you, but previous resolver logic basically set it as `sync` by default:
```
(schema, schemaOptions, resolverOptions = {}) =>
...
{ $async: resolverOptions && resolverOptions.mode === 'async' }
```

Can we consider it as bug and set default as `async`? Some users who inadvertently relied on it might have problems

elton-okawa:
Hi @jorisre, thanks for the kind words

`ajv` do have ways to transform data:
* [Ajv docs - Modifying data during validation](https://ajv.js.org/guide/modifying-data.html) - built in options to use for removing additional properties, assigning schema default values and coercing types
* [Ajv docs - User defined keywords](https://ajv.js.org/keywords.html) - allows you to define custom keywords to perform some transformation on specific fields
* [ajv-keywords](https://github.com/ajv-validator/ajv-keywords) - extension package from `ajv` maintainers with more keywords

At least for my feature request use case, those possibilities didn't work well mainly because I received a ready to use `ajv schema` to build a form and `keywords` that modify data must live in the schema. One possibility would be dynamically modify given schema but it didn't seem a good idea.

I'd like to understand why adding an optional `transform` option would be a breaking change, I tried to make sure that the behavior stayed the same if you didn't provide it:
```
const valid = validate(transform ? transform(values) : values);
```

ajv/src/ajv.ts
jorisre: Yes, we can consider it as a bug and set default as `async`

jorisre:
It would be a breaking change because the `transform` option should be is `schemaOptions` instead of `factoryOptions`. So:

```tsx
export type Resolver = <T extends FieldValues>(
  schema: Ajv.JSONSchemaType<T>,
  schemaOptions?: Ajv.Options, // ⬅️ you've to add a transform option here (like the classValidator resolver) but it'll introduce a breaking change
  factoryOptions?: { mode?: 'async' | 'sync' },
) => <TContext>(
  values: T,
  context: TContext | undefined,
  options: ResolverOptions<T>,
) => Promise<ResolverResult<T>>;
```

Ajv looks powerful but there are several ways to use it (with many plugins/modules) so it's complicated to create a resolver that supports all AJV's uses case.

I'm thinking to keep the resolver simple and you can your own.
At the same time, we can discuss, and iterate in order to find a better way to support data transformation that fits all the cases.

elton-okawa:
I see, now I fully understood why it'd be a breaking change, thanks!

I agree with you about keeping the resolver as simpler as possible because it might become a nightmare to maintain in the future.
Let's wait bluebill's opinion about adding the `transform`  option like `class-validator` does

bluebill1049:
Hi guys, sorry for the late reply. Could we hold the `transform` option? If this is a requirement, could we keep this a custom resolver at your app level?

elton-okawa:
Hi @bluebill1049 sure! We can hold the `transform` option changes without any problems.

Issue #549, opened 2023-04-22
Yup's resolver complains about schemas typed with Yup.Schema<MyFormType> in @hookform/resolvers v3.1.0 with yup v1.1.1


danielmarcano:
**Describe the bug**
When using `@hookform/resolvers` v3.1.0 and creating a `Yup` (v.1.1.1) schema, typing it as `Yup.Schema<MyFormType>` `yupResolver` complains about its type:

```ts
export type UserForm = {
  firstName?: string;
};

const formSchema: Yup.Schema<UserForm> = Yup.object().shape({
  firstName: Yup.string().optional()
});

const { control, reset } = useForm<UserForm>({
  mode: "all",
  resolver: yupResolver(formSchema),
});
```

**To Reproduce**

1. Install `yup` v1.1.1 and `@hookform/resolvers` v3.1.0 packages

2. Have TypeScript configured in your project

3. Create your form's type:

```ts
export type UserForm = {
  firstName?: string;
};
```

4. Create a `Yup` schema and type it as `Yup.Schema<MyFormType>`:

```ts
const formSchema: Yup.Schema<UserForm> = Yup.object().shape({
  firstName: Yup.string().optional()
});
```

5. Pass your form type to `useForm`, and `formSchema` to `yupResolver` within your `useForm`'s configuration object's `resolver` property:

```ts
const { control, reset } = useForm<UserForm>({
  mode: "all",
  resolver: yupResolver(formSchema)
});
```

6. Notice the error that `yupResolver` is showing regarding the passed `formSchema`

**Codesandbox link (Required)**

Here's a [codesandbox](https://codesandbox.io/s/focused-shaw-tkm3qs?file=/src/logic.ts:418-451) reproducing the issue.

**Expected behavior**

The `yupResolver` function should not complain about the given `formSchema`. It should consider it valid, and recognize its type, since this is considered valid code when using @hookform/resolvers v2.9.10 and yup v1.1.1.

**Screenshots**

<img width="943" alt="image" src="https://user-images.githubusercontent.com/93147810/233796641-744bb5f2-3d1b-47d2-9738-aecca81cf637.png">

**Desktop (please complete the following information):**

- OS: iOS 13.2.1
- Browser Chrome
- Version 112.0.5615.137

**Additional context**

[...]

PR #563, opened 2023-05-26
Fix schema type error mentioned in issue 549

henrikvolmer:
Fixes issue https://github.com/react-hook-form/resolvers/issues/549


bluebill1049:
Any chance you could add some test coverage on the fix? thank 🙏

henrikvolmer:
Do you have any suggestions, how to test a typescript type change? In my opinion, typescript will test it automatically on compile time.

bluebill1049:
If you can include usage in the test case would be good, that proof why the issue was there in the first place.

henrikvolmer:
Any news here? Do you need something?

jorisre (approved):
LGTM! thanks for the contribution

henrikvolmer:
@jorisre @bluebill1049 : I think this should have been better at least a minor release, since the type was `any` before and is now specialized. This is probably why some users have a problem with their types. Because of `any` the type check was disabled before and therefore wrong types didn't cause an error.

bluebill1049:
Thanks @henrikvolmer for supporting on this issue. Yes, I think a minor version would probably be the wiser release.

henrikvolmer:
@bluebill1049 : So how we move on? Can you change the patch version to a minor version?

jorisre:
We can:
- Update the release note
- Re-release a minor version with a release note

@bluebill1049 What do you prefer?

bluebill1049:
I think we just update the release note would be the better choice, cause the patch is already release. thanks @jorisre 🙏

jorisre:
Done https://github.com/react-hook-form/resolvers/releases/tag/v3.1.1

PR #584, opened 2023-06-25
feat: pass `options.names` as a second argument to a Vest suite

vonagam:
React Hook Form has a way to signal to a resolver for which fields validations were requested (`options.names`/`options.fields`) and Vest has a way to skip unrequested validations (`only`/`skip`/`include`). But currently that names information gets discarded and Vest cannot utilize that functionality. I think it makes sense to pass names as a second argument.

Note: `vestResolver` has a second argument which is and never was utilized, other adapters do not have such thing. I think it should be put on note somewhere for removal in next major release.

### DIFF
--- a/vest/src/__tests__/vest.ts
+++ b/vest/src/__tests__/vest.ts
@@ -68,4 +68,17 @@ describe('vestResolver', () => {
       ),
     ).toMatchSnapshot();
   });
+
+  it('should call a suite with values, validated field names and a context as arguments', async () => {
+    const suite = vi.fn(validationSuite) as any as typeof validationSuite;
+
+    await vestResolver(suite)(validData, { some: 'context' }, {
+      fields: { username: fields.username },
+      names: ['username'],
+      shouldUseNativeValidation,
+    });
+
+    expect(suite).toHaveBeenCalledTimes(1);
+    expect(suite).toHaveBeenCalledWith(validData, ['username'], { some: 'context' });
+  });
 });
--- a/vest/src/types.ts
+++ b/vest/src/types.ts
@@ -1,20 +1,23 @@
 import {
   FieldValues,
+  FieldName,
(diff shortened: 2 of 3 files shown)


jorisre:
Thanks for the contribution. Can you add an unit test for this new option?

vonagam:
Added a test with expected behavior.

jorisre:
Thanks for the test. 

The type should be updated as well: https://github.com/react-hook-form/resolvers/blob/master/vest/src/types.ts#L12

vonagam:
The linked line is one that I made note of. It is unused and can be removed, but this is a breaking change, I assume.

There is no change in types, because all arguments are the same. Currently vest suite signature is undefined. It is simply `ReturnType<typeof Vest.create>` so can be any function (even though implementation expects that first argument is field values, it is not expressed in types currently).

vonagam:
Ok, if it is allowed then it makes sense to type the expected call signature of a vest suite. I think this one is reasonable:
```ts
(values: TValues, names?: FieldName<TValues>[], context?: TContext) => void
```
So pass names as the second argument (such signature is shown many times in Vest documentation, mostly for passing to `only`) and a user supplied validation context as the third argument.

vonagam:
Any thoughts?

jorisre:
I tried your code locally, everything seems to work as expected without any breaking changes. 
@bluebill1049 Any thoughts? I'm about to merge this PR

bluebill1049:
Sounds good to me 👍 @jorisre

bluebill1049 (approved):

bluebill1049:
There is a PR here as well: https://github.com/react-hook-form/react-hook-form/pull/10590

jorisre:
FYI @bluebill1049 I merged but I'll release later with others fixes

PR #617, opened 2023-08-17
fix(class-validator): return instance unless rawValues is set

perebusquets:


class-validator/src/__tests__/class-validator.ts
```diff
@@ -145,3 +159,30 @@ it('validate data with validator option', async () => {
...
+it('should return from classValidatorResolver with `excludeExtraneousValues` set to true', async () => {
+  class SchemaTest {
+    @Expose()
+    @IsNumber({}, { message: `Must be a number`, always: true })
+    random: number;
+  }
+
+  const result = await classValidatorResolver(SchemaTest, {
+    transformer: {
+      excludeExtraneousValues: true,
```
jorisre, line 172: I don't find any documentation about `excludeExtraneousValues` and `extraneousField`. Do you have a link that explains these options?

jorisre (approved):
Thanks for your contribution!

PR #621, opened 2023-08-20
feat: add support for root errors for field array

jorisre:


bluebill1049 (dismissed):
Awesome!

src/toNestError.ts
```diff
@@ -5,6 +5,7 @@ import {
   Field,
```
bluebill1049, line 5: minor: maybe rename to `toNestErrors`?

jorisre: Done ✅

bluebill1049 (approved):

Issue #589, opened 2023-07-01
yupResolver() type mismatch with useForm() when required() on field with null default value


ypresto:
**Describe the bug**

Since v3.1.1, yupResolver() causes type mismatch between `required()` field and null value in defaultValues.

Probably caused by this PR: https://github.com/react-hook-form/resolvers/pull/563

@hookform/resolvers: v3.1.1 
react-hook-form: v7.45.1
yup: v1.2.0

**To Reproduce**

1. Prepare form with `useForm({ defaultValues, resolver: yupResolver(validationSchema) })`.
2. Add a null property (named `food` in codesandbox) to defaultValues (or just `defaultValues: { food: null }`).
3. Mark that property (`food`) as `string().required()` in validationSchema.
4. Type error at `useForm(...)`.

One example of type error:

```
Type 'Resolver<{ food: {}; }, any>' is not assignable to type 'Resolver<FormValues, any>'.
  Types of parameters 'values' and 'values' are incompatible.
    Type 'FormValues' is not assignable to type '{ food: {}; }'.
      Types of property 'food' are incompatible.
        Type 'FoodGroup | null' is not assignable to type '{}'.
          Type 'null' is not assignable to type '{}'.ts(2322)
```

**Codesandbox link (Required)**

https://codesandbox.io/s/react-hook-form-yupresolver-null-type-error-with-tfieldvalues-cd3tcc?file=/src/App.tsx:842-873

(Note: template for yupResolver is too old as it uses v6 of react-hook-form.)

**Expected behavior**
No type error found.

Note that react-select will set the value to null when it is cleared.

ivomasterche:
Hi, I have the same issue, and opened discussion here (still no replies)
https://github.com/orgs/react-hook-form/discussions/10618

Zerebokep:
I think my problem is also related: https://github.com/orgs/react-hook-form/discussions/10660

jorisre:
Related to #575

ivomasterche:
asegarra suggestion fixed the issue [https://github.com/react-hook-form/resolvers/issues/575#issuecomment-1693405121](url)
for me too. 

Adding NULL as a possible value for the input(select) in the FormInputType and specifying this type 
resolver: yupResolver<FormInputs>(schema),

was accepted and now there are no type errors.

jorisre:
[...]

Issue #575, opened 2023-06-13
Problem with yupResolver when one type in useForm generic is optional


arminhupka:
**Describe the bug**
When we provide optional key in useForm generic we getting error
`TS2322: Type 'Resolver<{ name: string | null | undefined; age: number; }>' is not assignable to type 'Resolver<{ name: string; age?: number | undefined; }, any>'.   Types of parameters 'values' and 'values' are incompatible.     Type '{ name: string; age?: number | undefined; }' is not assignable to type '{ name: string | null | undefined; age: number; }'.       Types of property 'age' are incompatible.         Type 'number | undefined' is not assignable to type 'number'.           Type 'undefined' is not assignable to type 'number'.` 

**To Reproduce**

- create new react project with vite
- create simple form wit useForm() and in generic add one optional key for example `age?: number`
- create yup schema with all details
- attach created schema to yupResolver
- we have a problem

**Codesandbox link (Required)**
https://codesandbox.io/s/quirky-fog-3695px?file=/src/TestView.tsx

**Expected behavior**
Works without problem

**Screenshots**
![image](https://github.com/react-hook-form/resolvers/assets/49194860/35b0cdc5-4917-473f-b280-d71715f4d324)

jorisre:
You don't need to manually type useForm from:

```tsx
const { register } = useForm<{ name: string; age?: number }>({
    resolver: yupResolver(form)
  });
```

to
```tsx
const { register } = useForm({
    resolver: yupResolver(form)
  });
```


Types are inferred from the schema

arminhupka:
@jorisre ok good but how i can now access to form values? I just need pass interface which looks like our schema to `SubmitHandler` or there is some different way to do that?

jorisre:
The schema is your single source of truth so

`SubmitHandler<Yup.InferType<typeof form>>`


https://github.com/jquense/yup#getting-started

jorisre:
Here is the updated release note, sorry for the mistake, it should be at least a minor version instead of a patch

https://github.com/react-hook-form/resolvers/releases/tag/v3.1.1

khuongtp:
I have a case that this auto infer feature seems not good at all. Before I can just pass the values and defaultValues conveniently like this 
![image](https://github.com/react-hook-form/resolvers/assets/62872729/4045d7a7-d552-4a84-82fa-2305399c3e50)
But now, TS keeps complaining, that's really annoying because we know that we pass somewhat the same type it expects
[...]

PR #625, opened 2023-08-28
fix(yup): resolve ts error when including optional key in yup schema

kotarella1110:
## Overview

This PR addresses the issue outlined in https://github.com/react-hook-form/resolvers/issues/575. The problem stems from inaccurate inference of generics (`TFieldValue`) in `yupResolver`.

For instance, when passing a schema with optional keys to `yupResolver`:

```tsx
const schema = yup.object({ test: yup.number() });

type TestValues = yup.InferType<typeof schema>;

function Form({ values }: { values: TestValues }) {
  const methods = useForm({
    values: values,
    resolver: yupResolver(testSchema)
  });
  return null;
}
```

The inferred type for `yupResolver` looks like this:

```ts
(alias) yupResolver<{
    test: number | undefined;
}>(schema: Lazy<any, yup.Maybe<yup.AnyObject>, any> | yup.ObjectSchema<{
    test: number | undefined;
}, yup.AnyObject, any, "">, schemaOptions?: yup.ValidateOptions<...> | ... 1 more ... | undefined, resolverOptions?: {
    ...;
}): Resolver<...>
import yupResolver
```

As you can see, the generics (`TFieldValue`) for `yupResolver` should have been inferred as optional keys, but they were inferred as required keys instead. This issue arises from the fact that the `ObjectSchema` in Yup extends the underlying schema's first argument using `MakeKeysOptional`. This utility type makes keys optional when the value of an object property is `undefined`. However, because this utility type is used, the expected inference isn't achieved, resulting in keys being mistakenly inferred as required.

https://github.com/jquense/yup/blob/1ee9b21c994b4293f3ab338119dc17ab2f4e284c/src/object.ts#L99

The incorrect inference of generics (`TFieldValue`) was causing problems when passing it as the form type to the Resolver. To resolve this, we fixed the return type by passing the correctly inferred schema type obtained through `Yup.InferType` as the argument to the Resolver.

## Resolved CSB

You can view the resolved example in the following CSB link: https://codesandbox.io/s/rhf-625-54jvv4

### DIFF
--- a/yup/src/yup.ts
+++ b/yup/src/yup.ts
@@ -43,7 +43,9 @@ const parseErrorSchema = (
 };
 
 export function yupResolver<TFieldValues extends FieldValues>(
-  schema: Yup.ObjectSchema<TFieldValues> | ReturnType<typeof Yup.lazy>,
+  schema:
+    | Yup.ObjectSchema<TFieldValues>
+    | ReturnType<typeof Yup.lazy<Yup.ObjectSchema<TFieldValues>>>,
   schemaOptions: Parameters<(typeof schema)['validate']>[1] = {},
   resolverOptions: {
     /**
@@ -56,7 +58,7 @@ export function yupResolver<TFieldValues extends FieldValues>(
      */
     raw?: boolean;
   } = {},
-): Resolver<TFieldValues> {
+): Resolver<Yup.InferType<typeof schema>> {
   return async (values, context, options) => {
     try {
       if (schemaOptions.context && process.env.NODE_ENV === 'development') {


kotarella1110:
This PR does not aim to resolve https://github.com/react-hook-form/resolvers/issues/589. However, regarding this issue, as mentioned in https://github.com/orgs/react-hook-form/discussions/10618#discussioncomment-6826058, I believe using the FormValue's values should be passed to yupResolver to address the inconsistency issue.

bluebill1049 (approved):
LGTM! Thanks @kotarella1110

jorisre:
Thanks a ton @kotarella1110

Issue #551, opened 2023-04-29
Inferring form keys and types from the zod return type


SeanCassiere:
**Is your feature request related to a problem? Please describe.**
It gets pretty annoying when using the `zodResolver`, that I need to infer the zod-schema type ***AND*** pass it into the `useForm` hook to get the necessary type-completions.
It's more frustrating as `zod` is known for its great TS-first approach and type inference (see the usage in tRPC).
This particular gap in the types is noticeable when using the `handleSubmit` in `rhf`, and all the key-values are possibly undefined, even though the zod-schema wouldn't allow for it to pass its validation.

**Describe the solution you'd like**
Using the provided zod-schema in the [`zodResolver`](https://github.com/react-hook-form/resolvers/blob/HEAD/zod/src/zod.ts#L58) to infer the [`TFieldValues`](https://github.com/react-hook-form/react-hook-form/blob/HEAD/src/types/form.ts#L100) used in the `useForm` hook in `rhf`.

**Describe alternatives you've considered**
Generics seem the only feasible way of doing this.

**Additional context**

PR #447, opened 2022-08-28
feat infer form type from zod schema and fix test

kagurazaka-0:
- feat infer FormType from zod schema without using type assertion
```ts
const HogeSchema = z.object({
  name: z.string(),
  age: z.number(),
});

const _ = useForm({ resolver: zodResolver(HogeSchema) });
// _ is UseFormReturn<{ name: string; age: number }>

_.watch("age") // work! it is number 
_.watch("name") // work! it is string
```

- fix type error in test

I referred to #417, thanks @maddijoyce !

PR #653, opened 2023-12-09
fix(zod): allow inferring specific `Resolver` type from `zodResolver`

### LINKED ISSUE #429: Infer types from zod schema
**Is your feature request related to a problem? Please describe.**
Zod automatically infers the typescript type from schema, so I declare my FieldValues like this:
```typescript
type FieldTypes = z.infer<typeof schema>
```
and then:
```typescript
const {...} = useForm<FieldTypes>({
  resolver: zodResolver(schema),
});
```
But I feel like the useForm hook should be abre to infer the type, since the useForm's resolver uses the generic TFieldValues `Resolver<TFieldValues, TContext>`

**Describe the solution you'd like**
The zodResolver should automatically infer TFieldValues given a zod schema

**Describe alternatives you've considered**
Using `useForm<z.infer<typeof schema>>` but I think this is ugly

mintchkin:
# Summary

Allows the result of `zodResolver` to infer the schema shape, based on the zod schema it's been passed. This matches the behavior of [the similar `yupResolver`](https://github.com/react-hook-form/resolvers/blob/3805d3eeae950ba6540f5d79d264c49287008d6a/yup/src/yup.ts#L61), as implemented in #563 

Resolves #429, #447, #551

### DIFF
--- a/zod/src/types.ts
+++ b/zod/src/types.ts
@@ -1,22 +1,4 @@
-import { FieldValues, ResolverResult, ResolverOptions } from 'react-hook-form';
-import { z } from 'zod';
+import type { zodResolver } from './zod';
 
-export type Resolver = <T extends z.Schema<any, any>>(
-  schema: T,
-  schemaOptions?: Partial<z.ParseParams>,
-  factoryOptions?: {
-    /**
-     * @default async
-     */
-    mode?: 'async' | 'sync';
-    /**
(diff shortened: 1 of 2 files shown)


dkAndFed:
Will this be merged at some point?

stephen776:
This would be really useful

PR #655, opened 2023-12-18
Valibot infinate loop in resolver

richardvanbergen:
**Describe the bug**
In recent versions of valibot they've introduced a `variant` method. This method does not set a `path` value in the issues array. Because of the way the `valibotResolver` loops over the issues returned from valibot, this can cause an infinite loop.

**To Reproduce**

1. Using the following schema:

```ts
const numberSchema = object({
  type: literal("number")
})

const stringSchema = object({
  type: literal("string")
})

const typeSchema = variant("type", [numberSchema, stringSchema])
```

2. Try to call the valibot resolver.

```ts
const data = {
  type: undefined
}

await valibotResolver(typeSchema, undefined, {
  mode: 'sync',
})(data, undefined, {
  fields,
  shouldUseNativeValidation,
});
```

3. Browser or test runner freezes.

**Additional context**

I've included tests in this PR. If you wish to demonstrate the issue you can revert the changes I made to `valibot.ts`.

Because there's no path returned from valibot we can't add the error to the resolver. I've requested a change in valibot to address this and the author [seems to agree](https://github.com/fabian-hiller/valibot/issues/303). This PR simply fixes the infinite loop and I will submit a second once the other issue has been addressed to update the dependency to the fixed version.

I've tried to follow the contributing guidelines as best as I can. It seems as they're out of date or were copied from somewhere else the following commands listed aren't in this repo as far as I can tell:

* tsd
* e2e
* start
* api-extractor

### DIFF
--- a/package.json
+++ b/package.json
@@ -249,7 +249,7 @@
     "superstruct": "^1.0.3",
     "typanion": "^3.14.0",
     "typescript": "^5.1.6",
-    "valibot": "^0.12.0",
+    "valibot": "^0.24.1",
     "vest": "^4.6.11",
     "vite": "^4.4.9",
     "vite-tsconfig-paths": "^4.2.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -126,8 +126,8 @@ devDependencies:
     specifier: ^5.1.6
     version: 5.1.6
   valibot:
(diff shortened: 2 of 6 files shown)


jorisre (approved):
Thanks for your contribution :)

Issue #629, opened 2023-09-06
Error Handling in useFieldArray with Root Error


KrabDr:
**Describe the bug**
Currently, there is an issue with error handling within useFieldArray, where errors associated with the root error are not functioning as expected.

**To Reproduce**

1. Go to [Codesandbox](https://codesandbox.io/s/yup-resolver-xjcg7z?file=/src/App.tsx)
2. Click on submit button
3. Observe that the error from the second field is displayed under the first field.

**Codesandbox link (Required)**
Sandbox demonstrating the issue: [Codesandbox](https://codesandbox.io/s/yup-resolver-xjcg7z?file=/src/App.tsx)

**Expected behavior**
Sandbox with the expected behavior: [Codesandbox](https://codesandbox.io/s/yup-resolver-manual-implementation-6r5w24?file=/src/App.tsx)

Upon investigation, it appears that the issue arises from the removal of empty values within an array before setting errors inside **toNestErrors**. When used within **Object.assign**, it results in errors being assigned to incorrect keys.

Versions

react-hook-form: 7.45.2
@hookform/resolvers: 3.3.1

PR #656, opened 2023-12-29
Fixes error handling for array errors with root error

SangJunBak:
Fixes https://github.com/react-hook-form/resolvers/issues/629

Issue:
If you had a field array with two elements and an error in the second element THEN a root error, `.compact` would recognize the second element error as an error in the first element.

Solution:
Get rid of `.compact`


jorisre (approved):
LGTM! thanks for your contribution

Yihao-G:
Do you mind releasing a new package version, please?

PR #670, opened 2024-03-22
Update README.md

william-lai-927:
From the Doc of valibot, the optional message is just apply when the type is not `string`, instead of using as 'required'. So the original is misleading. https://valibot.dev/api/string/


william-lai-927:
*P.S. the old format works on boolean & number, just not string. Probably version updated

PR #674, opened 2024-04-29
feat(typeboxResolver): make TypeBox resolver work with compiled schema

huynhducduy:
## Description

- Current implementation only makes use of the dynamic validation (aka `ValueCheck`) of `TypeBox` but not its `TypeCompiler` (a high-performance JIT) which is a killer feature, especially when working with recursive or constrained type validation.

## Summary of changes

- Add options for the user to pass a `TypeCheck` object (returned from `TypeCompiler.Compile()`) instead of plain schema.
- Add tests
- Add document for using `TypeCompiler`
- Fix typo in tests of the `ValueCheck` mode (it mentioned `zod` instead of `typebox`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@
   - [typanion](#typanion)
   - [Ajv](#ajv)
   - [TypeBox](#typebox)
+    - [With `ValueCheck`](#with-valuecheck)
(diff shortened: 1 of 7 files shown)


jorisre (approved):
LGTM, thanks!

PR #686, opened 2024-05-23
migrate arktype resolver to v2

ssalbdivad:
The `latest` version of `arktype` is now `2.0.0-dev.14`, which is consistent with an upcoming major release.

I've created a new major version of the `@hookform/resolvers/arktype` to match this.

I did have a question as to whether it was okay to resuse the existing `errors.byPath` property with a `type` alias for `code` as `FieldErrors`. It would have many additional properties users could introspect, so long as nowhere in hookform is there an assumption that there are no extra properties on that object (as can be seen by the snapshots with many additional properties).

If extra properties are problematic, I can adjust the code to just include `message` and `type` as they did before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -233,7 +233,7 @@
     "@vitejs/plugin-react": "^4.0.4",
     "ajv": "^8.12.0",
     "ajv-errors": "^3.0.0",
-    "arktype": "1.0.19-alpha",
+    "arktype": "2.0.0-dev.14",
     "check-export-map": "^1.3.0",
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.0",
--- a/arktype/package.json
+++ b/arktype/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@hookform/resolvers/arktype",
   "amdName": "hookformResolversArktype",
-  "version": "1.0.0",
+  "version": "2.0.0",
   "private": true,
   "description": "React Hook Form validation resolver: arktype",
   "main": "dist/arktype.js",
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -233,7 +233,7 @@
     "@vitejs/plugin-react": "^4.0.4",
     "ajv": "^8.12.0",
     "ajv-errors": "^3.0.0",
-    "arktype": "1.0.19-alpha",
```
bluebill1049, line 236: wow, we are using an `alpha`, that's not great, and now switching to `dev`. let's wait for the full release.

bluebill1049: thanks for the PR.

ssalbdivad: @bluebill1049 The integration isn't very deep, I can guarantee these parts of the API will be the same when a stable 2.0 is released.

In the meantime, giving users access to a resolver that is compatible with the latest version seems like a big win. Any reason not to merge this now?

ssalbdivad: @bluebill1049 Users have already asked to use 2.0 with `react-hook-form`. There will be stabilization period before there are no tags on the release that will not involve any of the APIs used in this PR.

Is there anything else I can do to get this merged?

ssalbdivad: @bluebill1049 Sorry to both you on this, just looking to understand if there is anything I can do to move this along.

bluebill1049 (approved):

ssalbdivad:
@bluebill1049 Thanks so much 🫶

bluebill1049:
Thank you & welcome @ssalbdivad 🙏

PR #695, opened 2024-06-25
chore: update valibot dependency to version >=0.33.0

fabian-hiller:
Unfortunately, I introduced a breaking change in a type of Valibot v0.33.0. So I updated the dependencies.


jorisre (approved):
Thanks!

PR #697, opened 2024-06-27
isZodError: check if error.errors is actually an array

### LINKED ISSUE #696: Zod Resolver Error: Cannot read properties of undefined (reading 'length')
**Describe the bug**
We have an error in our production server. (I have not gained clearance to share the stack trace) that indicates `errors` object can be undefined.

```
TypeError
Cannot read properties of undefined (reading 'length')
```

https://github.com/react-hook-form/resolvers/blob/f7a4fd46af51a2a732c10ed652f5abfa532f4198/zod/src/zod.ts#L13

This error appeared as soon as the user navigated to the page. The error cause is still yet unknown.

But we suspect that the `error.errors` object may be undefined and not just null:

https://github.com/react-hook-form/resolvers/blob/f7a4fd46af51a2a732c10ed652f5abfa532f4198/zod/src/zod.ts#L6

**To Reproduce**

My apologies, but we have not yet found a way to reliably reproduce this issue.

It may be caused by an imbalance between the versions of Zod and React Hook Form used in the project. However, for now, the error seems to originate from a single page and is not sporadic.

We are currently investigating further. For your reference, here are the versions of the relevant packages used in our project:

```
    "@hookform/resolvers": "3.5.0",
    "react-hook-form": "7.52.0",
    "zod": "3.19.0"
```

Include a codesandbox will help us to investigate the issue quicker.

- [Zod resolver](https://codesandbox.io/s/react-hook-form-zod-resolver-example-forked-ko5ll) Template

**Expected behavior**

We expected no errors happening when we open the page.

**Screenshots**

Not applicable

- OS: Windows 10
- Browser: Chrome
- Version: 26

**Additional context**

The zod form schema is a rather complex form of an object inside an object. The innermost object have string fields. Looks something like:

```ts
{
    o: {
        a: {
            x: string,
            y: string, 
        }
        b: {
            x: string,
            y: string, 
        }
    }
}
```

# Proposed Fix

image

Considering `isZodError` simply checks for the `errors` existence, should we make the function slightly more robust?

```ts
// check if it's actually an array, and not undefined nor null
const isZodError = (error: any): error is ZodError => Array.isArray(error?.errors);
```

ryanelian:
`ZodError` type expects `errors` prop to be an array

```ts
export declare class ZodError<T = any> extends Error {
    issues: ZodIssue[];
    get errors(): ZodIssue[];
    constructor(issues: ZodIssue[]);
    format(): ZodFormattedError<T>;
    format<U>(mapper: (issue: ZodIssue) => U): ZodFormattedError<T, U>;
    static create: (issues: ZodIssue[]) => ZodError<any>;
    toString(): string;
    get message(): string;
    get isEmpty(): boolean;
    addIssue: (sub: ZodIssue) => void;
    addIssues: (subs?: ZodIssue[]) => void;
    flatten(): typeToFlattenedError<T>;
    flatten<U>(mapper?: (issue: ZodIssue) => U): typeToFlattenedError<T, U>;
    get formErrors(): typeToFlattenedError<T, string>;
}
```

And `parseErrorSchema` function expects an array:

https://github.com/react-hook-form/resolvers/blob/f7a4fd46af51a2a732c10ed652f5abfa532f4198/zod/src/zod.ts#L8-L11

Recently, we encountered an issue where the props can be undefined for some reason. I propose this PR to add a stricter check to prevent unexpected circumstances where the `error.errors` object is not an array.

Resolve #696

### DIFF
--- a/zod/src/zod.ts
+++ b/zod/src/zod.ts
@@ -3,7 +3,7 @@ import { z, ZodError } from 'zod';
 import { toNestErrors, validateFieldsNatively } from '@hookform/resolvers';
 import type { Resolver } from './types';
 
-const isZodError = (error: any): error is ZodError => error.errors != null;
+const isZodError = (error: any): error is ZodError => Array.isArray(error?.errors);
 
 const parseErrorSchema = (
   zodErrors: z.ZodIssue[],


jorisre (approved):
Thanks for the fix!

PR #710, opened 2024-07-31
chore: upgrade @effect/schema

dudeofawesome:
`@effect/schema` was recently merged into the core `effect` package (indicating its API stability), so this PR drops the dependency on the old package and uses the core one, with a couple internal changes to align with the latest schema API.


dudeofawesome:
Closing in favor of #720

PR #714, opened 2024-09-09
perf(ajv): reuse compiled schema

matthiasschwarz:
Moved initialization of Ajv, ajvErrors and the creation of the adjusted schema outside of the Resolver to allow the instance of Ajv to reuse the compiled schema on multiple invocations of the Resolver. Also adjusted the example usage in the README to show that the ajvResolver should only be called once for the schema.

This aligns with the recommended approach of compiling a schema only once as described in the docs under [Managing schemas](https://ajv.js.org/guide/managing-schemas.html).

Overall this should also significantly reduce the memory footprint since the instance of Ajv and the modified schema is reused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,14 +465,14 @@ const schema = {
   additionalProperties: false,
 };
 
+const resolver = ajvResolver(schema);
+
 const App = () => {
   const {
     register,
     handleSubmit,
     formState: { errors },
-  } = useForm({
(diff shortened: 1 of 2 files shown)


jorisre:
Thank you for your contribution. With your changes, it is no longer possible to dynamically modify the `schemaOptions` and `resolverOptions`. Although this is likely not a common use case, it does introduce a breaking change.

PR #717, opened 2024-10-04
chore: update valibot dependency to v1

fabian-hiller:
I've updated the Valibot dependency for the React Hook Form resolver to v1. We've changed the type signature, so this change is necessary. All other changes in our v1 beta version are only internal.

---

An union of schema library authors (Zod, Valibot, ArkType, ...) collaborating on a standard interface for schema libraries called [Standard Schema](https://github.com/standard-schema/standard-schema). This simplifies implementation and prevents vendor lock-in. So far Valibot v1 supports this spec and Zod will follow soon. In the long run, the explicit implementation of a Valibot and Zod resolver may no longer be necessary.


bluebill1049 (approved):
LGTM

bluebill1049:
@fabian-hiller there is no breaking change with this v1 correct?

fabian-hiller:
There is. But only the type signature. So users will see a TS error if they try to use Valibot v1 beta with the old adapter.

PR #737, opened 2025-01-30
feat(effectResolver): returns either all errors or only the first one based on `criteriaMode`

### LINKED ISSUE #721: [effect-ts] Only the last issue per field is reported
**Describe the bug**
If the filed is a union, only the last issue is reported. This can be cumbersome especially with optional fields, which are just unions with undefined.

**To Reproduce**
Create a Schema which will produce multiple `ParseIssues` for a single filed

https://codesandbox.io/p/sandbox/react-hook-form-validationschema-v6-forked-nwp2d2

**Expected behavior**
The error should contain the error "Please enter a valid phone number in international format." when submit is clicked.

jorisre:
Fixes #721 

_(do not merge)_


bluebill1049 (approved):
LGTM!

PR #739, opened 2025-02-05
feat: automatically infer values from schema

jorisre:
This PR ensures consistency across resolvers by automatically inferring values from the schema when applicable, as we already do for the Yup resolver.

Additionally, it adds a small comparison table in the README to highlight the supported features of each resolver.


package.json
```diff
@@ -318,5 +318,8 @@
   },
   "peerDependencies": {
     "react-hook-form": "^7.0.0"
+  },
+  "dependencies": {
+    "caniuse-lite": "^1.0.30001698"
```
bluebill1049, line 323: we need this?

bluebill1049 (approved):

package.json
nozaq: Is this comment resolved? I couldn't find any reference to this package in the diff.

PR #749, opened 2025-02-24
fix(zod): update return type in zodResolver to use z.input instead of…

### LINKED ISSUE #743: Zod resolver infered types are incorect since 4.1.0
**Describe the bug**
zodResolver now infer schema type which is great but unfortunately produces invalid typings

**To Reproduce**
Setup a project using zod, typescript and react-hook-form, create a simple form such as

```
const FORM_SCHEMA = z.object({
  s: z.string(),
});

const App = () => {
  const form = useForm({
    resolver: zodResolver(FORM_SCHEMA),
  });

  const [s] = form.watch(["s"]);

  console.log(s);
  ...
};
```

https://codesandbox.io/p/sandbox/wispy-wildflower-fk7ymp?file=%2Fsrc%2Findex.js%3A11%2C4

**Expected behavior**
`s` is typed as `string` but should be typed as `string | undefined`

**Screenshots**
Image

- OS: macOS 15.3
- Browser any
- Versions: zod 3.24.2, react-hook-form 7.54.2, @hookform/resolvers 4.1.0

**Additional context**

I personally used to provide `useForm` both TFieldValues and TTransformedValues typings so that everything is properly typed such as
```
const form = useForm<
    DeepPartial<z.input<typeof SCHEMA>>,
    unknown,
    z.output<typeof SCHEMA>
  >({
    resolver: zodResolver(SCHEMA),
  });

```

but even this is now impossible since types are clashing: `Types of parameters 'values' and 'values' are incompatible`

Thank you 🙇

diegotraid:
Resolves #743 

For the function zodResolver to return a type compatible with react-hook-form, it must return `Resolver<z.input<typeof schema>>` instead of `Resolver<z.infer<typeof schema>>`.

### DIFF
--- a/zod/src/__tests__/Form.tsx
+++ b/zod/src/__tests__/Form.tsx
@@ -10,17 +10,22 @@ const schema = z.object({
   password: z.string().nonempty({ message: 'password field is required' }),
 });
 
-type FormData = z.infer<typeof schema> & { unusedProperty: string };
+type SchemaInput = z.input<typeof schema>;
+type FormData = z.output<typeof schema> & { unusedProperty: string };
 
 function TestComponent({
   onSubmit,
-}: { onSubmit: (data: z.infer<typeof schema>) => void }) {
+}: { onSubmit: (data: z.output<typeof schema>) => void }) {
   const {
     register,
     handleSubmit,
     formState: { errors },
-  } = useForm({
+  } = useForm<SchemaInput, unknown, FormData>({
     resolver: zodResolver(schema), // Useful to check TypeScript regressions
+    defaultValues: {
+      username: '',
+      password: '',
+    },
   });
 
   return (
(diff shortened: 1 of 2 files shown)


controversial:
Is it possible for `useForm` to also infer the `TTransformedValues` parameter from the Resolver? or can `useForm` only pick up the `TFieldValues` *input* shape from the Resolver?

diegotraid:
> Is it possible for `useForm` to also infer the `TTransformedValues` parameter from the Resolver? or can `useForm` only pick up the `TFieldValues` _input_ shape from the Resolver?

That, I think, it's the main issue here.
I think resolvers doesn't take transformed values into consideration:

`export type Resolver<TFieldValues extends FieldValues = FieldValues, TContext = any>`

While useForm does:
`useForm<TFieldValues extends FieldValues = FieldValues, TContext = any, TTransformedValues extends FieldValues | undefined = undefined>`

controversial:
At the very least, this PR will be an improvement over the current situation

diegotraid:
Waiting for https://github.com/react-hook-form/react-hook-form/pull/11951 to be merged into master to provide a more correct solution to this.

controversial:
That PR was merged into react-hook-form’s `main` 3 days ago I believe

diegotraid:
Closing since this is being resolved with PR #753

PR #753, opened 2025-02-26
feat: implement typed Input/Output interface for resolvers

### LINKED ISSUE #743: Zod resolver infered types are incorect since 4.1.0

jorisre:
This PR implements typed Input/Output interfaces for resolvers, which needs https://github.com/react-hook-form/react-hook-form/pull/12638

Next steps will involve extending this implementation to all applicable resolver modules.

Note: This PR will fail until the release of https://github.com/react-hook-form/react-hook-form/pull/12638

Fixes #743


zod/src/__tests__/zod.ts
```diff
@@ -89,4 +91,86 @@ describe('zodResolver', () => {
...
+
+    expectTypeOf(resolver).toEqualTypeOf<
+      Resolver<{ id: string }, any, { id: string }>
+    >();
+  });
+
+  it('should correctly infer the output type from a zod schema when a different input type is specified (only input type is specified)', () => {
+    const resolver = zodResolver<{ id: string }>(z.object({ id: z.string() }));
+
+    expectTypeOf(resolver).toEqualTypeOf<
```
jorisre, line 120: @kotarella1110 Confused why this test is failing while the previous one works fine. Could this be a TypeScript limitation?

kotarella1110: This is a limitation of TypeScript's partial inference behavior—once you explicitly specify one or more generic parameters, TypeScript no longer infers the rest automatically.

For more details, please refer to the following issue: https://github.com/microsoft/TypeScript/issues/26242

As a result, the default value of the `Schema` generic in `zodResolver`, which is `z.output<typeof schema>`, is passed as the third argument to `Resolver`.  

Therefore, modifying the test as follows will make it pass:  

```ts
expectTypeOf(resolver).toEqualTypeOf<
  Resolver<{ id: string }, any, z.output<z.ZodSchema<any, any, any>>>
>();
```

kotarella1110: Considering the limitation of TypeScript's type inference, a good solution is to avoid specifying default values for the third generic parameter. This way, when generic parameters are explicitly provided, we can enforce that the third argument (`Output`) is required.

Here’s the solution:

```ts
export function zodResolver<
  Input extends FieldValues,
  Context,
  Output,
  Schema extends z.ZodSchema<any, any, any> = z.ZodSchema<any, any, any>
>(
  schema: Schema,
  schemaOptions?: Partial<z.ParseParams>,
  resolverOptions: {
    mode?: "async" | "sync";
    raw?: boolean;
  } = {}
): Resolver<
  Input,
  Context,
  unknown extends Output ? z.output<typeof schema> : Output
> {
  return {} as any;
}
```

With this change:

```ts
zodResolver(z.object({ id: z.number() })); // No type error
zodResolver<{ id: string }>({ id: z.number() }); // Type error
zodResolver< // No type error
    { id: string },
    { context: any },
    { id: boolean }
  >(z.object({ id: z.string() }));
```

zod/src/__tests__/zod.ts
```diff
@@ -89,4 +91,86 @@ describe('zodResolver', () => {
...
+      SubmitHandler<{
+        id: string;
+      }>
+    >();
+  });
+
+  it('should correctly infer the output type from a Zod schema when different input and output types are specified for the handleSubmit function in useForm', () => {
+    const resolver = zodResolver<{ id: string }, any, { id: boolean }>(
+      z.object({ id: z.number() }),
+    );
```
kotarella1110, line 164: It seems there is a simple mistake here. should be:

```suggestion
    const resolver = zodResolver<{ id: string }, any, { id: boolean }>(
      z.object({ id: z.string() }),
    );
```

Additionally, since the `Input` type of the generic parameters and the `Schema` type don't match, a type error should actually occur here. This would indicate that the code is not type-safe.  

To fix this, you should pass the `Output` and `Input` types to the `ZodSchema` like so:

```ts
export function zodResolver<
  Input extends FieldValues,
  Context,
  Output,
  Schema extends z.ZodSchema<Output, any, Input> = z.ZodSchema<Output, any, Input>
>(
  //...
);
```

zod/src/__tests__/zod.ts
kotarella1110: I believe the following change, including https://github.com/react-hook-form/resolvers/pull/753#discussion_r1973619256, would be the best approach:

```ts
export function zodResolver<
  Input extends FieldValues,
  Context,
  Output,
  Schema extends z.ZodSchema<Output, any, Input> = z.ZodSchema<Output, any, Input>
>(
  schema: Schema,
  schemaOptions?: Partial<z.ParseParams>,
  resolverOptions: {
    mode?: "async" | "sync";
    raw?: boolean;
  } = {}
): Resolver<
  Input,
  Context,
  unknown extends Output ? z.output<typeof schema> : Output
> {
  return {} as any;
}
```

zod/src/__tests__/zod.ts
bluebill1049: thanks @kotarella1110 and 👋

jorisre: Fixed ✅

zod/src/__tests__/zod.ts
jorisre: Hi @kotarella1110 , I've implemented the majority of your suggested changes. However, I wasn't able to apply the `z.ZodSchema<Output, any, Input>` recommendation - specifically the Input type portion. This appears to be incompatible with Zod's internal type system. I don't believe this is a significant issue for our implementation.

zod/src/zod.ts
```diff
@@ -80,14 +80,27 @@ function parseErrorSchema(
...
   schemaOptions?: Partial<z.ParseParams>,
   resolverOptions: {
     mode?: 'async' | 'sync';
     raw?: boolean;
   } = {},
-): Resolver<z.infer<typeof schema>> {
+): Resolver<
+  Input,
+  Context,
+  (typeof resolverOptions)['raw'] extends true
```
jorisre, line 98: @kotarella1110  I'm encountering a TypeScript error with our recent changes due to the `raw` option. We should theoretically be able to infer the input type when `raw: true` is specified.

I attempted to implement this inference, but it resulted in the following error: 

<details>
 <summary>Show error</summary>
 
 
 > Type '(values: Input, _: Context | undefined, options: ResolverOptions<Input>) => Promise<{ errors: FieldErrors; values: ({} & Input) | Awaited<...>; } | { ...; }>' is not assignable to type 'Resolver<Input, Context, unknown extends Output ? output<Schema> : Output>'.
  Type 'Promise<{ errors: FieldErrors; values: ({} & Input) | Awaited<Output>; } | { values: {}; errors: FieldErrors<Input>; }>' is not assignable to type 'ResolverResult<Input, unknown extends Output ? output<Schema> : Output> | Promise<ResolverResult<Input, unknown extends Output ? output<...> : Output>>'.
    Type 'Promise<{ errors: FieldErrors; values: ({} & Input) | Awaited<Output>; } | { values: {}; errors: FieldErrors<Input>; }>' is not assignable to type 'Promise<ResolverResult<Input, unknown extends Output ? output<Schema> : Output>>'.
      Type '{ errors: FieldErrors; values: ({} & Input) | Awaited<Output>; } | { values: {}; errors: FieldErrors<Input>; }' is not assignable to type 'ResolverResult<Input, unknown extends Output ? output<Schema> : Output>'.
        Type '{ errors: FieldErrors; values: ({} & Input) | Awaited<Output>; }' is not assignable to type 'ResolverResult<Input, unknown extends Output ? output<Schema> : Output>'.
          Type '{ errors: FieldErrors; values: ({} & Input) | Awaited<Output>; }' is not assignable to type 'ResolverError<Input>'.
            Types of property 'values' are incompatible.
              Type '({} & Input) | Awaited<Output>' is not assignable to type '{}'.
                Type 'Awaited<Output>' is not assignable to type '{}'.
                  Type 'Output | (Output extends object & { then(onfulfilled: infer F, ...args: infer _): any; } ? F extends (value: infer V, ...args: infer _) => any ? Awaited<V> : never : Output)' is not assignable to type '{}'.
                    Type '(null | undefined) & Output' is not assignable to type '{}'.
                      Type 'undefined & Output' is not assignable to type '{}'.ts(2322)
</details>

zod/src/__tests__/zod.ts
kotarella1110: @jorisre 

Could you share what specific type incompatibility you encountered when trying to apply `z.ZodSchema<Output, any, Input>`? I'd like to understand the issue better.

With `z.ZodSchema<Output, any, any>`, the schema and the expected input type can become misaligned, leading to potential runtime errors that TypeScript should ideally catch. For example, in the following case, `id` should be a `string` in `Input`, but the Zod schema allows a `number`, and TypeScript wouldn't detect this issue:

```ts
zodResolver<{ id: string }, unknown, { id: string }>(
  z.object({ id: z.number() }).transform(({ id }) => {
    return { id: String(id) };
  })
);
```

jorisre: Sure! Here is an example: 

```tsx
const schema = z.object({ id: z.number() });
const resolver = zodResolver<{ id: string }, any, z.output<typeof schema>>(
  schema,
);
```

This produces the following type error:

> Argument of type 'ZodObject<{ id: ZodNumber; }, "strip", ZodTypeAny, { id: number; }, { id: number; }>' is not assignable to parameter of type 'ZodType<{ id: number; }, any, { id: string; }>'.
  The types of '_input.id' are incompatible between these types.
    Type 'number' is not assignable to type 'string'.

jorisre: You're right, that was a misunderstanding on my part :) Everything's clear now. I've submitted an updated test using `transform` that makes more sense.

zod/src/zod.ts
jorisre: I'm still encountering this issue and am unsure how to resolve it. Any help from the community would be greatly appreciated :)

Screenshot 2025-03-06 at 10 06 34

controversial: Hi @jorisre!

First, I believe you’re going to have trouble getting type checking to work with types that are conditional on `resolverOptions.raw` until https://github.com/microsoft/TypeScript/pull/56941 re-lands (as https://github.com/microsoft/TypeScript/pull/61359) in the next upcoming version of TypeScript

Before that, I’d recommend using an overload to capture the separate return type for `raw: true`.

---

Second, I believe one thing making it hard for types to pass is that the generics have too many degrees of freedom. If you have:
```ts
export function zodResolver<
  Input extends FieldValues,
  Context,
  Output,
  Schema extends z.ZodSchema<Output, any, Input> = z.ZodSchema<
    Output,
    any,
    Input
  >
>(
  schema: Schema
```
it’s possible for users to instantiate the Schema parameter with a more specific input/output type than `Input`/`Output`, which will make it harder for the types to pass within your function. Instead, I’d recommend something like the following:
```ts
export function zodResolver<
  Input extends FieldValues,
  Context,
  Output,
>(
  schema: z.ZodSchema<Output, any, Input>,
```
By removing the separate type parameter for Schema (which is redundant with the existing type parameter anyway) there‘s a lot less room for TypeScript errors in the family of “___ could be instantiated with a different subtype of constraint ___”

controversial:
@jorisre I believe I made some progress on the types here, in #755 and in https://github.com/react-hook-form/react-hook-form/pull/12638

Let me know what you think!

zod/src/zod.ts
kotarella1110: @controversial 

You're absolutely right. It makes sense, and it looks like we don’t really need to use generics here.
Thanks!

@jorisre 

Based on @controversial 's feedback, I'd like to propose the following changes:

```ts
export function zodResolver<Input extends FieldValues, Context, Output>(
  schema: z.ZodSchema<Output, any, Input>,
  schemaOptions?: Partial<z.ParseParams>,
  resolverOptions?: {
    mode?: "async" | "sync";
    raw: true;
  }
): Resolver<Input, Context, Input>;
export function zodResolver<Input extends FieldValues, Context, Output>(
  schema: z.ZodSchema<Output, any, Input>,
  schemaOptions?: Partial<z.ParseParams>,
  resolverOptions?: {
    mode?: "async" | "sync";
    raw?: false;
  }
): Resolver<Input, Context, Output>;
export function zodResolver<Input extends FieldValues, Context, Output>(
  schema: z.ZodSchema<Output, any, Input>,
  schemaOptions?: Partial<z.ParseParams>,
  resolverOptions: {
    mode?: "async" | "sync";
    raw?: boolean;
  } = {}
): Resolver<Input, Context, Output> {
  // ...
}
```

kotarella1110: Oops, I just realized that @controversial already created a https://github.com/react-hook-form/resolvers/pull/755 with the same changes. 😅
Sorry for the duplicate comment, and thanks for the PR!

jorisre:
I need some help fixing this TS issue. @bluebill1049  @controversial  or @kotarella1110 

Screenshot 2025-03-26 at 11 11 00

kotarella1110:
@jorisre 

> I need some help fixing this TS issue.

I've created a PR. Please feel free to merge it if you like it.

https://github.com/react-hook-form/resolvers/pull/757

kotarella1110 (approved):
LGTM! Appreciate your hard work!

jorisre:
> LGTM! Appreciate your hard work!

Appreciate your help, thanks again!

bluebill1049 (approved):
Thanks everyone! Huge effort @jorisre

bluebill1049 (approved):

zod/package.json
```diff
@@ -11,7 +11,7 @@
   "types": "dist/index.d.ts",
   "license": "MIT",
   "peerDependencies": {
-    "react-hook-form": "^7.0.0",
+    "react-hook-form": "7.55.0",
```
spiftire, line 14: Should this be locked to version 7.55.0?
```suggestion
    "react-hook-form": "^7.55.0",
```

jorisre: Good catch @spiftire

arktype/src/arktype.ts
huynhducduy, line 1: Why do we switch from `arktype` native implementation to `standard-schema`? And why not all standard-schema-compatible resolvers follow this approach?

huynhducduy: Also, using this approach will lead to the resolver failing to infer the type from the schema, unless the user specifically installs `@standard-schema/spec` as their dependency

huynhducduy: apologize for mentioning you guys @jorisre @bluebill1049 @kotarella1110

jorisre: We switched to standard-schema from the native arktype implementation because type inference in Arktype wasn't working as expected. But if you've found a solid solution, we're happy to review a PR :)

PR #757, opened 2025-03-26
fix(typebox): improve type interface

kotarella1110:
resolve https://github.com/react-hook-form/resolvers/pull/753#issuecomment-2753877072


jorisre (approved):
Thanks for the help and the explanation :)

Issue #747, opened 2025-02-23
Standard Schema Resolver `Could not resolve "@standard-schema/utils"`


ryanagillie:
**Describe the bug**
Standard schema resolver fails to import `@standard-schema/utils`

**To Reproduce**

1. Go to https://stackblitz.com/edit/vitejs-vite-neec4edf?file=src%2FApp.tsx
2. Check console

**Expected behavior**
Have the standard schema resolver not have an import error on `@standard-schema/utils`

**Screenshots**
![Image](https://github.com/user-attachments/assets/14217b4a-d81f-4e3c-b693-465ba058b52f)

ryanagillie:
Gonna cc @EskiMojo14 as this started happening with the most recent fix for it

EskiMojo14:
okay, so the difference with the standard-schema resolver vs the individual library resolvers is that the user likely won't already have `@standard-schema/spec` and `@standard-schema/utils` installed. 

Possible solutions:
- make sure all users have them installed by including them as explicit dependencies
- document somewhere that `standardSchemaResolver` users also need to install them manually
  - if doing this, they should still probably be optional peer dependencies to ensure compatible versions.

EskiMojo14:
once installed, reproduction builds successfully: https://stackblitz.com/edit/vitejs-vite-es3gsqwe?file=package.json

graup:
> make sure all users have them installed by including them as explicit dependencies
> this means they get installed for everyone, including people who won't use standardSchemaResolver

I think this is a pretty ugly solution. Is there now better way?

EskiMojo14:
> I think this is a pretty ugly solution. Is there no better way?
Making it an optional peer dependency, which is also pretty ugly.

to be frank, the end goal of Standard Schema is that the concept of a "resolver" is no longer necessary, since all compliant libraries speak the same language.

I don't know if there are specific benefits at the moment to using a specific resolver for your schema library of choice (and I'm sure there are some supported that don't currently support Standard Schema), but *eventually*™ RHF should be able to remove the need for a resolver, and everybody uses standard schema.

graup:
Thanks for the context. I thought Standard Schema is just another validation library, but if it's a standard API going to used everywhere, I guess it makes sense to have that as a general dependency!

PR #765, opened 2025-04-09
add prebuild script to collect peerdeps

### LINKED ISSUE #762: issue: Peer dependencies not picked up by package managers
### Version Number

5.0.1

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/hkjpmt

### Steps to reproduce

- open codesandbox vm
- cd apps/site
- pnpm ls --depth 2 --prod

Observe

```
@hookform/resolvers 5.0.1
├── @standard-schema/utils 0.3.0
└─┬ react-hook-form 7.55.0 peer
  └── react 19.0.0 peer
```

### Expected behaviour

For the resolvers to build reliably, the package manager needs to know what the peer deps are so that they may link them properly. Without this information, you can end up with unpredictable build errors where, by pure coincidence, a package (which is not listed in the dependencies such as effect) is not available. I am running into this error but cannot replicate it on codesandbox since the preconditions are dependent on a complicated combination of the layout of your workspace, hoisting, package manager, and indeed the build tool you are using (in my case it is nextjs + turbopack).

Listing peer dependencies only in the package.json files in the sub-folder (effect in this case) does not supply enough information to the package manager to be able to hoist / link reliably. They need to be at the top level.

The only solution is to break this into individually-adressable packages (@hookform/resolver-effect for example) or to list all the peer deps in the top level package.

_No response_

### Relevant log output

```shell
➜  site git:(main) ✗ pnpm ls --depth=2 --prod
Legend: production dependency, optional only, dev only

react-vite-ts@0.0.0 /project/workspace/apps/site (PRIVATE)

dependencies:
@hookform/resolvers 5.0.1
├── @standard-schema/utils 0.3.0
└─┬ react-hook-form 7.55.0 peer
  └── react 19.0.0 peer
effect 3.14.6
├── @standard-schema/spec 1.0.0
└─┬ fast-check 3.23.2
  └── pure-rand 6.1.0
effect-example link:../../packages/effect
└─┬ effect 3.14.6
  ├── @standard-schema/spec 1.0.0
  └── fast-check 3.23.2
react 19.0.0
react-dom 19.0.0
├── react 19.0.0 peer
└── scheduler 0.25.0
react-hook-form 7.55.0
└── react 19.0.0 peer
```

### Code of Conduct

arlyon:
This adds a simple js script that is run before builds and commits that collects all peerDependencies from all subfolders together and inserts them into the root package.json. This closes #762

### DIFF
--- a/package.json
+++ b/package.json
@@ -221,6 +221,7 @@
     "build:vine": "microbundle --cwd vine --globals @hookform/resolvers=hookformResolvers,react-hook-form=ReactHookForm,@vinejs/vine=vine",
     "build:fluentvalidation-ts": "microbundle --cwd fluentvalidation-ts --globals @hookform/resolvers=hookformResolvers,react-hook-form=ReactHookForm",
     "build:standard-schema": "microbundle --cwd standard-schema --globals @hookform/resolvers=hookformResolvers,react-hook-form=ReactHookForm,@standard-schema/spec=standardSchema",
+    "prebuild": "node ./config/collect-peerdeps.js",
     "postbuild": "node ./config/node-13-exports.js && check-export-map",
     "lint": "bunx @biomejs/biome check --write --vcs-use-ignore-file=true .",
     "lint:types": "tsc",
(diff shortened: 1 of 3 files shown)


arlyon:
The question of #747 pops to mind here. If the official stance of this library is that there is a hard dependency on standard schemas with the view of eventually making it the _only_ option then great.

package.json
```diff
@@ -317,9 +318,91 @@
     "zod": "^3.24.2"
   },
   "peerDependencies": {
-    "react-hook-form": "^7.55.0"
+    "react-hook-form": "^7.55.0",
+    "@hookform/resolvers": "^2.0.0",
```
aqeelat, line 322: why are you depending on the package itself?

aqeelat (commented):
This is how we can make sure we're not depending on ourselves.

Issue #768, opened 2025-04-13
Zod 4 support


channyeintun:
**Need support for Zod 4 and @zod/mini**
Zod 4 beta and @zod/mini beta are out recently. Currently, resolver doesn't work with Zod 4 and @zod/mini.

hamdani-hash:
I have added related discussion in [React Hook Form Repo](https://github.com/orgs/react-hook-form/discussions/12744). Looking forward for the recent update to this...

semet:
I have added related discussion in [React Hook Form Repo](https://github.com/orgs/react-hook-form/discussions/12744). Looking forward for the recent update to this...

wenerme:
xref https://github.com/colinhacks/zod/pull/4074

ukuq:
I found a solution and intend to submit a PR, but I noticed the version number on GitHub is much lower than the one on npm. Not sure why - maybe the author hasn't open-sourced the latest code? Who knows.
Below is my solution:
```
Object.defineProperty(ZodError.prototype, "errors", {
    get() {
        return this.issues;
    },
    enumerable: true,
    configurable: true,
});

const form = useForm<{ [x: string]: never }>({
    resolver: zodResolver(FormSchema),
    defaultValues: data as { [x: string]: never },
});
```

ukuq:
> I found a solution and intend to submit a PR, but I noticed the version number on GitHub is much lower than the one on npm. Not sure why - maybe the author hasn't open-sourced the latest code? Who knows. Below is my solution:
> 
> ```
> Object.defineProperty(ZodError.prototype, "errors", {
>     get() {
>         return this.issues;
>     },
>     enumerable: true,
>     configurable: true,
> });
> 
> const form = useForm<{ [x: string]: never }>({
>     resolver: zodResolver(FormSchema),
>     defaultValues: data as { [x: string]: never },
> });
> ```

It seems this is controversial, so I'll close the PR for now.

raptoria:
sorry, a bit of a n00b here. What is the standard schema resolver? Can you provide a link?
Does this solve the issue of seeing an Uncaught in Promise error when my form first loads (because validation fails)? My  expectation was that @hookform/resolvers/zod already supported and used safeParse by default so I'm confused as to why I see a promise error

igomonteiro:
@raptoria can you try something like this?

`const form = useForm<z.infer<typeof SignInSchema>>({
    resolver: standardSchemaResolver(SignInSchema),
    defaultValues: {
[...]

PR #776, opened 2025-05-31
feat: zod v4 support #768

alexcraviotto:
## Support for Zod v4 with `zodResolverV4`

This PR adds support for the new version of Zod (v4) by introducing a new resolver: `zodResolverV4`. This allows users to work with schemas written using the latest version of Zod while maintaining compatibility with older implementations.

To ensure backward compatibility, the original `zodResolver` has **not** been removed or modified.

Due to issues related to tests in `typeschema.ts`, I’ve temporarily installed **two separate versions of Zod** (`zod` and `zod-v4`) to avoid breaking existing functionality. Ideally, in the future, it would be great to update those parts to also use the latest version of Zod.

### Usage

Using the new resolver is simple. You just need to import and use `zodResolverV4` like so:

```ts
const { control, handleSubmit, formState: { errors } } = useForm<User>({
  resolver: zodResolverV4(UserSchema),
  defaultValues: {
    name: '',
    device: {
      platform: 'ios',
      versions: []
    }
  }
});

```
Big thanks to @hukpo for mentioning this solution in [Issue #768](https://github.com/react-hook-form/resolvers/issues/768) 🙌

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,6 +5,7 @@
       "name": "@hookform/resolvers",
       "dependencies": {
         "@standard-schema/utils": "^0.3.0",
+        "zod-v4": "npm:zod@^3.25.0",
       },
       "devDependencies": {
         "@sinclair/typebox": "^0.34.30",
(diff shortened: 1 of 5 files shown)


zod/src/zodv4.ts
```diff
@@ -0,0 +1,143 @@
...
+import {
+  FieldError,
+  FieldErrors,
+  FieldValues,
+  Resolver,
+  ResolverError,
+  ResolverSuccess,
+  appendErrors,
+} from 'react-hook-form';
+import { ZodError, z } from 'zod-v4/v4';
```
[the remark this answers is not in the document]
alexcraviotto: It needs to be done that way to work.

zod/src/index.ts
```diff
@@ -1 +1,2 @@
-export * from './zod';
+export { zodResolver } from './zod';
```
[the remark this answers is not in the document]
alexcraviotto: We do it this way to be able to export both resolvers of both versions.

bluebill1049:
thanks @colinhacks for the review 🙏

bun.lock
```diff
@@ -5,6 +5,7 @@
       "name": "@hookform/resolvers",
       "dependencies": {
         "@standard-schema/utils": "^0.3.0",
+        "zod-v4": "npm:zod@^3.25.0",
```
CHC383, line 8: The lock file also needs to be updated after moving zod-v4 to devDependencies

zod/src/zodv4.ts
```diff
@@ -0,0 +1,134 @@
...
+import {
+  FieldError,
+  FieldErrors,
+  FieldValues,
+  Resolver,
+  ResolverError,
+  ResolverSuccess,
+  appendErrors,
+} from 'react-hook-form';
+import { ZodError } from 'zod-v4/v4';
```
CHC383, line 11: I think for v4-mini compatibility, the code should use `core.$ZodError` instead.

https://zod.dev/packages/core?id=errors#errors

zod/src/zodv4.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+export function zodResolver<Input extends FieldValues, Context, Output>(
+  schema: core.$ZodType<Output, Input>,
+  schemaOptions?: Partial<core.ParseContext<core.$ZodIssue>>,
+  resolverOptions?: {
+    mode?: 'async' | 'sync';
+    raw?: false;
+  },
+): Resolver<Input, Context, Output>;
```
controversial, line 76: it seems like the overload for `raw: true` was incorrectly removed, which breaks the types when this option is passed @alexcraviotto

controversial: see https://github.com/react-hook-form/resolvers/pull/776#discussion_r2122258830

jorisre:
Thanks a lot @alexcraviotto for your contribution. Closing in favor of #777

PR #777, opened 2025-06-05
feat: support Zod 4

colinhacks:
Per some discussions I've been having with @bluebill1049, this PR adds support for Zod 4.

The other PR by @alexcraviotto (https://github.com/react-hook-form/resolvers/pull/776) is really solid but we decided to go another direction.

 
- This PR bumps the minimum Zod version to `3.25.0`
- With this change, it;s possible to support both Zod 3 and Zod 4 in a single `zodResolver` (via Zod's [subpath versioning](https://github.com/colinhacks/zod/issues/4371) scheme)
- I've designed this to continue working with Zod versions pre-3.25.0, but it's on a best-effort basis. For instance, you won't get perfect type signatures on parse params (the second argument to `zodResolver`). There were zero breaking changes between Zod 3.24 and Zod 3.25 so there's no reason not to upgrade.

Both Zod 3 and Zod 4 schemas can be passed into `zodResolver()`:

```ts
import { zodResolver } from "@hookform/resolvers";
import * as z3 from "zod/v3"; // or just "zod"
import * as z4 from "zod/v4";

const oldResolver = zodResolver(z3.object({ name: z.string() }));
const newResolver = zodResolver(z4.object({ name: z.string() }));
```

### DIFF
--- a/zod/src/zod.ts
+++ b/zod/src/zod.ts
@@ -143,7 +143,11 @@ type NonRawResolverOptions = {
 interface Zod3Type<O = unknown, I = unknown> {
   _output: O;
   _input: I;
+  _def: {
+    typeName: string;
+  };
 }
+
 interface Zod4Type<O = unknown, I = unknown> {
   _zod: {
     output: O;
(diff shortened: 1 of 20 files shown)


zod/src/zod.ts
```diff
@@ -63,79 +79,237 @@ function parseErrorSchema(
   return errors;
 }
 
+function parseZod4Issues(
+  zodErrors: z4.$ZodIssue[],
+  validateAllFieldCriteria: boolean,
+) {
+  const errors: Record<string, FieldError> = {};
+  // const _zodErrors = zodErrors as z4.$ZodISsue; //
+  for (; zodErrors.length; ) {
```
bluebill1049, line 88: Interesting syntax, I guess it's the same as `while`, wonder is it just perf benefit?

zod/src/zod.ts
```diff
@@ -63,79 +79,237 @@ function parseErrorSchema(
...
+            ),
+          } satisfies ResolverError<Input>;
+        }
+
+        throw error;
+      }
+    };
+  }
+
+  throw new Error('Invalid input: not a Zod schema');
```
bluebill1049, line 314: aha, nice!

bluebill1049 (commented):
LGTM

Jamess-Lucass:
I've noticed the error messages returned from the form errors has changed when referencing zod/v4.

```typescript
import { useForm } from "react-hook-form";
import { zodResolver } from "@hookform/resolvers/zod";
import { z } from "zod";

const formSchema = z.object({
  firstname: z.string().max(2),
});

type FormData = z.infer<typeof formSchema>;

function App() {
  const {
    register,
    handleSubmit,
    formState: { errors, isSubmitting },
  } = useForm<FormData>({
    resolver: zodResolver(formSchema),
  });

  const onSubmit = (data: FormData) => {
    console.log("Form data:", data);
  };

  return (
    <div>
      <form onSubmit={handleSubmit(onSubmit)}>
        <label>First Name:</label>
        <input id="firstname" type="text" {...register("firstname")} />

        <button type="submit" disabled={isSubmitting}>
          Submit
        </button>
      </form>

      <p>{errors.firstname?.message}</p>
    </div>
  );
}

export default App;
```

The the input "longer_than_2_chars"

`errors.firstname?.message` would return "String must contain at most 2 character(s)", which i would expect.

when updating the import to `import { z } from "zod/v4";`

`errors.firstname?.message` now just returns "Invalid input". It doesn't seem to get the underlying zod error message anymore.

jorisre (approved):
Thanks for the black TypeScript magic 🧙‍♂️ Really appreciate you jumping in and helping out with this. Everything looks solid and comes with a good DX too 🙌

juliansommer:
@colinhacks I ran into issues when trying to upgrade to zod/v4 now. In nextjs getting some stacktrace error when it attempts to handle validation checking on submit. Will open issue if needed

MartinCura:
> Hello, I'm trying to use react-hook-form with zod/v4. It seems to me that this code should work, yet I'm facing an issue about incompatible types between the one expected by zodResolver and the passed schema.
> 
> ```ts
> //"react-hook-form": "^7.57.0",
> import { useForm } from "react-hook-form";
> 
> //"@hookform/resolvers": "^5.1.0",
> import { zodResolver } from "@hookform/resolvers/zod";
> 
> //"zod": "^3.25.57",
> import { z } from "zod/v4";
> 
> const loginSchema = z.object({
> 	email: z.email("email.invalid"),
> 	password: z
> 		.string()
> 		.nonempty("password.required")
> 		.min(8, "password.minLength"),
> });
> export const Component = (): void => {
> 	useForm({
> 		resolver: zodResolver(loginSchema),
> 		mode: "onChange",
> 		defaultValues: {
> 			email: "",
> 			password: "",
> 		},
> 	});
> };
> ```
> 
> image
> 
> Does anyone have any idea of why this happens? Thanks

Found the same thing, not sure why. I switched to using the `standardSchemaResolver` and it seems to work so i worked around it 🤷🏼 Not sure if you found the reason.

sergiojoker11:
Backwards compatibility got broken with this change. In reality, this has been a 'feat!'

PR #798, opened 2025-07-25
fix: zod v4 peer deps

theoludwig:
As explained in the "Library authors" part of the release of Zod v4, we can support easily Zod v4 without any code changes except updating the peer dependency version range (as the work was already done in #777).

Ref: https://github.com/colinhacks/zod/releases/tag/v4.0.1


jorisre:
Thanks!

PR #824, opened 2025-10-22
fix: fix nesting issue due to the errors order

volodymyr-kushnir:
This fixes a nesting issue I've originally hit with the ajv resolver (albeit the issue itself is applicable to all resolvers) which is that `toNestErrors` may lose keys depending on the order of the keys returned by the `parseErrorSchema`, e.g. in this particular example the error for the `config` is actually last (the console output of `parsedErrors` is sorted thus a bit misleading) and simply replaces the `config: { type: …, userId: … }` object that existed one nesting step earlier —

image

I was explaining this to my coworkers so I have a screencast (100 seconds) that illustrates the issue, attaching it here in hope that it'd explain the changes behind the PR better than my writing —

https://github.com/user-attachments/assets/2e5eb038-3e73-434b-bb49-1a54a30747cd

Also I was not very comfortable with passing the objects straight to `Object.assign`, I think I've hit a bug with objects being mutated in place, so even though I can no longer reproduce it I've still wrapped "foreign" objects in a `structuredClone`, just to be on a safe side.

### DIFF
--- a/src/__tests__/__snapshots__/toNestErrors.ts.snap
+++ b/src/__tests__/__snapshots__/toNestErrors.ts.snap
@@ -3,6 +3,11 @@
 exports[`transforms flat object to nested object 1`] = `
 {
   "name": {
+    "firstName": {
+      "message": "third message",
+      "ref": undefined,
+      "type": "rd",
+    },
     "message": "first message",
     "ref": {
       "reportValidity": [MockFunction spy],
@@ -25,6 +30,11 @@ exports[`transforms flat object to nested object 1`] = `
 exports[`transforms flat object to nested object and shouldUseNativeValidation: true 1`] = `
 {
   "name": {
+    "firstName": {
+      "message": "third message",
+      "ref": undefined,
+      "type": "rd",
+    },
     "message": "first message",
     "ref": {
       "reportValidity": [MockFunction spy] {
--- a/src/__tests__/toNestErrors.ts
+++ b/src/__tests__/toNestErrors.ts
@@ -3,6 +3,7 @@ import { toNestErrors } from '../toNestErrors';
 
 const flatObject: Record<string, FieldError> = {
   name: { type: 'st', message: 'first message' },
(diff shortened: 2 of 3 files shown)


volodymyr-kushnir:
Fair enough, I've removed `structuredClone` due to `"target"` being `"ES2017"` and left it like it was originally. If I manage to reproduce the issue I was once observing I'll circle back to this. I've tried to quickly plumb it with the `JSON.parse(JSON.stringify(…))` technique but an array of say `[empty × 3, {…}]` is getting stringified as `[null,null,null,{}]` which doesn't play well down the road (resulting nested object gets `"0":null` key-value pair(s), etc.).

volodymyr-kushnir:
Hi, could you please clarify why this PR was closed? I don't see a review comment or explanation, and this doesn't look like it was superseded — [#848](https://github.com/volodymyr-kushnir/resolvers/issues/848) (merged) fixed a different bug (the regex-metacharacter `SyntaxError` in `escapeForRegex`), not the nesting/ordering issue this PR addresses.

I just re-verified against current master (827f20b) and the bug still reproduces:
```
import { toNestErrors } from './src/toNestErrors';

const errors = {
  'config.type': { type: 'invalid_type', message: 'type is wrong' },
  'config.userId': { type: 'invalid_type', message: 'userId is wrong' },
  config: { type: 'custom', message: 'config is invalid overall' },
};

toNestErrors(errors, { fields: {}, shouldUseNativeValidation: false });
// => { config: { type: 'custom', message: 'config is invalid overall' } }
// `type` and `userId` are silently dropped
```
Since `set(fieldErrors, path, error)` overwrites whatever was previously nested at that path, any error keys that happen to iterate before a shorter parent-path key get wiped out. Happy to rebase/update this PR (e.g. drop the `structuredClone` usage per the earlier feedback and find another way to avoid the in-place mutation concern) if there's an alternative approach you'd prefer — just let me know what didn't work about this one.

PR #699, opened 2024-07-02
feat: add typeschema resolver

jorisre:
This is the follow-up to the PR on #663  by [decs](https://github.com/decs) with merge conflicts resolved. Huge thanks to [decs](https://github.com/decs)

Bonus: 
- Update all dependencies to latest version and fix errors/warnings
- Migrate from prettier/eslint to biomejs ⚡ 
- Migrate from PNPM to Bun ⚡ 
- Migrate from lint-staged/husky to lefthook ⚡ 
- Fix build warnings
- Reduce bundle size for:
   - Arktype and might fix this issue #693 
   - Typeschema
   - Vine
   - Valibot

PR #837, opened 2026-01-28
fix(tests): fix all failing tests

### LINKED ISSUE #832: Many tests failing
I tried to run the tests with `bun run test` and ran into several issues. Initially, the tests would not run and Bun showed "No test files found, exiting with code 1". This was fixed by removing `'/\\.'` from the `vitest.config.ts`.

Once I got the tests running, I noticed that many are failing. Here's the output:

>Test Files  17 failed | 49 passed
Tests  44 failed | 213 passed

```
FAIL  arktype/src/__tests__/arktype.ts > arktypeResolver > should return a single error from arktypeResolver when validation fails
Error: Snapshot `arktypeResolver > should return a single error from arktypeResolver when validation fails 1` mismatched

- Expected
+ Received

  {
    "errors": {
      "accessToken": {
        "message": "accessToken must be a number or a string (was missing)",
        "ref": undefined,
-       "type": "required",
+       "type": "",
      },
      "birthYear": {
        "message": "birthYear must be a number (was a string)",
        "ref": undefined,
-       "type": "domain",
+       "type": "",
      },
...
```

[This PR](https://github.com/react-hook-form/resolvers/commit/18e423f670de6eddc3ca550e7505f6a24b469aad) changed the GitHub workflow to no longer run the tests in CI.

BrendanC23:
1. Remove `/\\.` from `vitest.config.ts`. This was necessary to fix the error `No test files found`. This can also be fixed by updating Vitest to version 4, but this has a change to how snapshots are stored. See [the migration guide](https://vitest.dev/guide/migration.html#changes-to-mocking). All of the tests are found by removing `/\\.`, so the Vitest upgrade is better in a later PR.
1. Inlined `z` in `vitest.config.ts` to spy on `parse()` from `zod/v4/core`. Without this, a `TypeError: Cannot redefine property: parse` error occurs.
1. Change `bun test` to `bun run test` in CONTRIBUTING.md. `bun test` runs Bun's native test runner, not the `test` script.
1. Add `Test` stage to `.github/workflows/main.yml`. It was removed in [this PR](https://github.com/react-hook-form/resolvers/pull/699) and the tests haven't been running in CI.
1. Fix failing tests

Notes:

1. All
    * Fixed tests that called `useForm` outside of a component. This fixed the errors `Error: Invalid hook call. Hooks can only be called inside of the body of a function component.` and `TypeError: Cannot read properties of null (reading 'useRef')`.
1. Arktype
    * [This PR](https://github.com/react-hook-form/resolvers/pull/753) changed from directly calling Arktype to using standard schema. Part of this transition was changing `type: error.code` to `type: ''` (because `error.code` is not part of the standard schema). This required an update to the snapshot.
1. Superstruct
    * Fix validation of schemas that perform coercion (`{ coerce: true }` must be passed in)
    * Added additional overload for type definition `superstructResolver` that improves handling of coerced schemas (This fixes a type error with the form values).
    * Fix definition of coercion schema so that it correctly converts from number to string
    * As mentioned in [this issue](https://github.com/ianstormtaylor/superstruct/issues/1159), there is no way to infer the input type of a coerced schema. A [separate PR](https://github.com/ianstormtaylor/superstruct/pull/1181/) would be necessary. Some of the type inference tests have been changed to manually provide the input type (this enables `watch` to return the correct value for a coerced schema).
1. Typeschema
    * The `for` loop was looping through the `schemaErrors` variable, but the current error was based on the `typeschemaErrors[0]` variable. This caused only the first error to be returned.
    * Change from spying on `typeschema.validate` to watching `schema["~standard"]`
1. Zod
    * Fixed Zod tests that spied on `schema.parse()` instead of `z4.parse()`
    * Added missing snapshots

Fixes https://github.com/react-hook-form/resolvers/issues/832

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,9 @@ jobs:
 
       - name: Lint
         run: bun lint
+      
+      - name: Test
+        run: bun run test
           
       - name: TS
         run: bun lint:types
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ Here is a quick guide to doing code contributions to the library.
 
 4. If you've added a code that should be tested, ensure the test suite still passes.
 
-   > bun test
+   > bun run test
 
 5. Try to write some unit tests to cover as much of your code as possible.
 
--- a/arktype/src/__tests__/__snapshots__/arktype.ts.snap
+++ b/arktype/src/__tests__/__snapshots__/arktype.ts.snap
@@ -6,41 +6,41 @@ exports[`arktypeResolver > should return a single error from arktypeResolver whe
     "accessToken": {
       "message": "accessToken must be a number or a string (was missing)",
       "ref": undefined,
-      "type": "required",
+      "type": "",
     },
     "birthYear": {
       "message": "birthYear must be a number (was a string)",
       "ref": undefined,
-      "type": "domain",
+      "type": "",
     },
     "dateStr": {
       "message": "dateStr must be a Date (was missing)",
       "ref": undefined,
-      "type": "required",
+      "type": "",
     },
     "email": {
       "message": "email must be an email address (was "")",
       "ref": {
         "name": "email",
       },
-      "type": "pattern",
+      "type": "",
     },
     "enabled": {
       "message": "enabled must be boolean (was missing)",
       "ref": undefined,
-      "type": "required",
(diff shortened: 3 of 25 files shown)


bluebill1049:
sorry i missed this pr could you merge the latest?

bluebill1049 (approved):

PR #849, opened 2026-06-15
fix: preserve standard schema root errors

### LINKED ISSUE #841: issue: Improper handling of errors with empty path in Standard Schema
### Version Number

5.2.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/currying-wildflower-h472k9

### Steps to reproduce

1. Use the standard schema resolver with a validator that may return an error with either no path or an empty path (in my case, typia when union discrimination fails uses an empty path)
2. Submit the form
3. Notice how despite a validation fail occured, no errors are raised and handleSubmit calls the submit handler with an empty object

It is caused by the code correctly going down [the "validation failure" code path](https://github.com/react-hook-form/resolvers/blob/02286db88310fd23e1de61ad880c23e210801813/standard-schema/src/standard-schema.ts#L96-L106), and returning `{ values: {}, errors: parseErrorSchema() }`, but [`parseErrorSchema()`](https://github.com/react-hook-form/resolvers/blob/02286db88310fd23e1de61ad880c23e210801813/standard-schema/src/standard-schema.ts#L6-L33) starts from an empty object, iteratively adding errors to it _while skipping empty paths_, causing the returned error to end up as an empty object, which signals no error to react-hook-form.

### Expected behaviour

Given that [the path of a standard schema issue is explicitly optional in the spec](https://github.com/standard-schema/standard-schema/blob/5d9e4d0bc23e2fc33415fe054d620b7cacb70b0f/packages/spec/src/index.ts#L81), I would expect the standard schema resolver to properly handle such issue types.

Even if issues without a path are considered non-supported, at the very least, it should not return an empty object as a validated result, as that breaks the type contract of handleSubmit.

Firefox

### Relevant log output

```shell

```

### Code of Conduct

pupuking723:
Fixes #841

## Proposed Changes

- Preserve Standard Schema issues that do not provide a path by mapping them to `errors.root`
- Keep `criteriaMode: "all"` behavior for multiple root-level issues
- Add regression coverage for missing-path Standard Schema issues

Verification:

- `/Users/mac/.bun/bin/bun test ./standard-schema/src/__tests__/standard-schema.ts -t root`
- `/Users/mac/.bun/bin/bun lint`
- `/Users/mac/.bun/bin/bun run lint:types`
- `/Users/mac/.bun/bin/bun run build:standard-schema`
- `git diff --check`

### DIFF
--- a/standard-schema/src/standard-schema.ts
+++ b/standard-schema/src/standard-schema.ts
@@ -3,25 +3,53 @@ import { StandardSchemaV1 } from '@standard-schema/spec';
 import { getDotPath } from '@standard-schema/utils';
 import { FieldError, FieldValues, Resolver } from 'react-hook-form';
 
+const schemaRootKey = Symbol('Schema root');
+
+function appendErrorMessage(error: FieldError, message: string | undefined) {
+  if (!message) {
+    return;
+  }
+
+  error.message = error.message ? `${error.message}\n${message}` : message;
+}
+
+function appendErrorTypes(error: FieldError, types: FieldError['types']) {
+  if (!types) {
+    return;
+  }
+
+  const nextTypes = error.types || {};
+
+  for (const message of Object.values(types)) {
+    nextTypes[Object.keys(nextTypes).length] = message;
+  }
+
+  error.types = nextTypes;
+}
+
 function parseErrorSchema(
   issues: readonly StandardSchemaV1.Issue[],
   validateAllFieldCriteria: boolean,
 ) {
-  const errors: Record<string, FieldError> = {};
+  const errors = new Map<string | typeof schemaRootKey, FieldError>();
 
   for (let i = 0; i < issues.length; i++) {
     const error = issues[i];
-    const path = getDotPath(error);
+    const path = getDotPath(error) || schemaRootKey;
+    const fieldError = errors.get(path);
 
-    if (path) {
-      if (!errors[path]) {
-        errors[path] = { message: error.message, type: '' };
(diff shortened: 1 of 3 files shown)


standard-schema/src/standard-schema.ts
```diff
@@ -11,21 +11,19 @@ function parseErrorSchema(
 
   for (let i = 0; i < issues.length; i++) {
     const error = issues[i];
-    const path = getDotPath(error);
+    const path = getDotPath(error) || 'root';
```
evelynesideofdux: This would conflict with a key on the object called `root`, would a [`Symbol("Schema root")`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) maybe work here?

standard-schema/src/standard-schema.ts
```diff
@@ -11,21 +11,19 @@ function parseErrorSchema(
...
 
-    if (path) {
-      if (!errors[path]) {
-        errors[path] = { message: error.message, type: '' };
-      }
+    if (!errors[path]) {
+      errors[path] = { message: error.message, type: '' };
+    }
```
evelynesideofdux: While we're at it, maybe we should avoid silencing additional errors on the same path, something like

```suggestion
    if (errors[path]) {
      errors[path].message = `${errors[path].message}\n${error.message}`;
    } else {
      errors[path] = { message: error.message, type: '' };
    }
```

Issue #828, opened 2025-11-26
issue: TypeScript errors with @hookform/resolvers when Zod is not declared as peer dependency


drac94:
### Version Number

5.2.2

### Codesandbox/Expo snack

N/A - Monorepo-specific issue with pnpm workspaces

### Steps to reproduce

1. Set up a pnpm workspace with multiple apps using different Zod versions
2. Install @hookform/resolvers@5.2.2 in App A with Zod 4.x
3. Install @hookform/resolvers@3.9.1 in App B with Zod 3.x  
4. Use pnpm overrides to enforce specific Zod versions per app
5. Run TypeScript typecheck
6. Observe TypeScript errors about incompatible Zod types even though runtime dependency resolution is correct

### Expected behaviour

@hookform/resolvers@5.2.2 should declare `zod` as an optional peer dependency in package.json, similar to how other validation libraries are declared (yup, joi, etc.).

This would allow package managers like pnpm to properly symlink the correct Zod version into @hookform/resolvers' node_modules, ensuring TypeScript can resolve to the correct type definitions.

Expected package.json to include:
```json
"peerDependencies": {
  "react-hook-form": "^7.55.0",
  "zod": "*"
},
"peerDependenciesMeta": {
  "zod": {
    "optional": true
  }
}
```

**Actual behaviour (what went wrong):**

@hookform/resolvers@5.2.2 does not declare `zod` as a peer dependency. While it uses `@standard-schema/utils` to provide validation, TypeScript still needs to resolve Zod types when using `zodResolver`.

In a pnpm workspace with multiple Zod versions:
- pnpm doesn't create a Zod symlink in @hookform/resolvers' node_modules
- TypeScript walks up the directory tree to find Zod types
- TypeScript finds the wrong Zod version (e.g., finds Zod 3.x when it should use Zod 4.x)
- This causes type errors like: `Type 'import(".../zod@4.0.17/...")' is not assignable to type 'import(".../zod@3.25.76/...")'`

**Workaround:**
Had to manually add peer dependency using pnpm's `packageExtensions`:
```json
"pnpm": {
  "packageExtensions": {
    "@hookform/resolvers@5.2.2": {
      "peerDependencies": {
        "zod": "*"
      }
    }
  }
}
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

SalahAdDin:
@bluebill1049 This happens yet. Can you tell us which versions solve this error?

Issue #814, opened 2025-08-26
issue: misconfigured peerDependencies in internal packages


ItamarGronich:
### Version Number

5.2.1

### Codesandbox/Expo snack

Explained in the description

### Steps to reproduce

# Description

Currently the structure of this package and repo is that there's the main package at the root of the repo, and then a directory with a private package.json file for each resolver:

```text
/
|_ package.json { /* all package publish configuration*/ }
  |_ zod
  |  |_ package.json { private: true, peerDependencies: { /* zod|etc... */} }
  |_ yup
  |  |_ package.json { private: true, peerDependencies: { /* ... */} }
  |_ etc...
```

So the entire thing is published as one package, while the internal packages are essentially ignored by package managers.

This means that their manifests and dependency requirements are essentially invisible to package managers. Because they aren't _real_ packages.

This means that:

1. package managers won't autoinstall peers because they don't know they exist
2. no warning for missing or wrong peer dependencies
3. package managers with have strict peer dependency resolution won't error out on wrong peers.
4. in some package managers, the peer can be resolved from the root of the workspace because all packages are hoisted, but in pnpm for example you can set `hoist: false` to prevent this behavior. and if you combine that with pnpm workspaces, then zod would not be accessible to `@hookform/resolvers`. Thats because pnpm creates virutal sandboxes for all peer libraries, so even if zod IS installed in the workspace project, it still won't be found. You can reproduce that by checking out the `pnpm-unhoisted` branch on my reproductino repo.

## Steps to reproduce

1. Clone the reproduction: repository https://github.com/ItamarGronich/repro-resolvers
2. Follow the `README.md` in the repo root for reproduction steps.

## Suggested solution

One possible solution to this issue is to restructure the package layout with workspaces.

Each resolver would be it's own package which is also published (not private).
That would also have benefits for your users because they can install a smaller more minial package with less dependencies.

How this would be structure is like this

1. Every resolver has it's own package
2. Code in the root src directory would be moved into a `@hookforms/resolvers-core` package
3. For backwards compatibility, the old package would be kept as a wrapper around the new packages, re-exporting their functionality the same way it does today but instead of the code existing in the package, it would be imported and re-exported from the new packages.
4. Modify the docs to reflect the new package structure and provide guidance on how to migrate to the new setup.

### Expected behaviour

peerDependencies are resolved correctly for all package manager with every setting.

### Relevant log output

```shell
[...]

PR #850, opened 2026-07-08
fix: declare validation libraries as optional peerDependencies

### LINKED ISSUE #762: issue: Peer dependencies not picked up by package managers

colinhacks:
## Problem

Each resolver entry point (`@hookform/resolvers/zod`, `/yup`, `/joi`, …) imports its validation library at runtime, but the root `package.json` lists those libraries only under `devDependencies`. They are declared as neither dependencies nor peer dependencies of the published package, so package managers can't see the requirement — it resolves only by hoisting luck. Strict installs (pnpm, Yarn PnP) and dependency-graph tooling flag or miss them. The per-adapter `package.json` files do declare peers, but they are `"private": true` and aren't wired into `exports`, so no resolver ever reads them.

Closes #762. Refs #765, #814, #828.

## Change

Declare every adapter's validation library as an optional peer dependency of the root package — `peerDependencies` plus `peerDependenciesMeta.<lib>.optional: true`. `react-hook-form` stays a required peer, and `@standard-schema/utils` stays a runtime dependency.

This takes the goal of #765 but keeps the version ranges hand-authored rather than deriving them from the adapter manifests. Derivation silently dropped any library whose nested manifest never declared it — most visibly `zod`, and still `yup`, `joi`, and `computed-types`, which import their library but list no peer for it. Derivation also pinned wrong ranges: several nested manifests carry a caret that excludes the version the repo actually tests against (`@sinclair/typebox` `^0.25.24` vs. the tested `0.34`; `@vinejs/vine` `^2` vs. the tested `3`; `nope-validator` `^0.12` vs. the tested `1.x`).

The `zod` entry point imports both `zod/v3` and `zod/v4/core`, so its range spans both majors: `^3.25.0 || ^4.0.0` (those subpaths first shipped in zod 3.25.0).

`config/check-peer-deps.js` guards completeness. Run in `postbuild` alongside `check-export-map`, it scans each adapter's imports and fails the build if one reaches for a library that the root doesn't declare as a peer — so a newly added adapter can't ship a phantom dependency, while the ranges stay hand-controlled.

Optional peer dependencies are the established pattern for a package with pluggable integrations — ESLint plugin presets, TypeORM database drivers.

## Verification

- `bun run build` — green; `postbuild` runs the new check (`✓ peer deps: all 20 resolver adapters declare their validation libraries`).
- `bun install --frozen-lockfile` — green (lockfile updated to match).
- `biome check` — clean.


colinhacks:
The peer dependency versions are intentionally hand-written. The prebuild script only checks for *existence*. This gives the maintainers more control in cases like Zod and Valibot, where the supported version ranges are a bit more  complicated.

bluebill1049 (approved):

kelnei:
Heads-up that this looks like it made the optional peer set unsatisfiable as of 5.4.1 — any resolution that reaches 5.4.1+ fails with ERESOLVE, including for consumers that only use `zodResolver` and have neither valibot nor typeschema anywhere in their tree.

Minimal repro (npm 12.0.1, Node 24.18.0) — a four-line manifest:

```json
{
  "name": "repro",
  "version": "1.0.0",
  "dependencies": {
    "@hookform/resolvers": "5.4.2",
    "react-hook-form": "^7.82.0",
    "zod": "^4.4.3"
  }
}
```

```
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error Found: valibot@0.39.0
npm error   peerOptional valibot@"^0.39.0" from @typeschema/valibot@0.14.0
npm error     peerOptional @typeschema/valibot@"0.14.0" from @typeschema/main@0.14.1
npm error       peerOptional @typeschema/main@">=0.13.7" from @hookform/resolvers@5.4.2
npm error Could not resolve dependency:
npm error peerOptional valibot@"^1.0.0 || ^1.0.0-beta.4 || ^1.0.0-rc" from @hookform/resolvers@5.4.2
```

The two entries can't both hold: `valibot@^1` on one side, and `@typeschema/main@>=0.13.7` on the other — whose only satisfying chain is 0.14.1 → `@typeschema/valibot@0.14.0` (an exact pin) → `valibot@^0.39.0`. npm materializes the optional-peer chain and the two valibot constraints then collide. `@typeschema` appears unmaintained (still 0.14.x, pinned to valibot 0.39 while valibot itself is at 1.4.2), so an upstream bump won't reconcile the two.

@colinhacks — re: the ranges being intentionally hand-written for cases like Zod and Valibot, that makes sense; this seems to be the case where the hand-written pair is jointly unsatisfiable no matter what the consumer installs, since one of the peers transitively constrains the other.

Either of these would resolve it as far as I can tell: drop `@typeschema/main` from the peer list, or widen valibot to `^0.39.0 || ^1.0.0 || ^1.0.0-beta.4 || ^1.0.0-rc`. Pinning `5.4.0` exactly works as a consumer-side workaround meanwhile.

(I'd have opened an issue, but blank issues are disabled and the bug template requires a Codesandbox link, which can't demonstrate an npm resolution failure. Happy to move this wherever you'd prefer it.)

PR #873, opened 2026-08-12
support joi v18 as a peer dependency

sulaysumaria:
- Widens the `joi` peer dependency range to `^17.0.0 || ^18.0.0` so projects using joi 18 aren't blocked by the resolver's peer dependency constraint.
- Bumps the `joi` devDependency to `^18.2.3` so the test suite actually runs and is verified against joi 18, not just joi 17.
- This is purely a dependency range fix. All existing joi resolver tests pass unmodified against joi 18.
- joi 17 support is kept in the peer range (rather than being replaced) to avoid a breaking change for existing consumers who haven't upgraded yet.


sulaysumaria:
@bluebill1049 , Can you check this?

bluebill1049 (approved):

